PR #6, opened 2017-09-24
Support for printing version

jinankjain:
This may work but could not verify it because I don't have any local kubeconfig. May be you can test it.


jinankjain:
@klautcomputing I merged both the PRs over here. So now the new command to find version is `kubeaudit version` so now we don't even need the flag

PR #8, opened 2017-09-25
Version 2

jinankjain:


jinankjain:
So now the command is `kubeaudit version`

PR #12, opened 2017-09-29
The "kubectl image" command requires the "-i" or "--image" flags.

marc-barry:
The `Example usage` for the `kubectl image` command is incorrect. The `-i` or `--flags` are required.


klautcomputing:
Double checked the README. It's fine.

PR #13, opened 2017-09-29
Print INFO message for matching image:tag

marc-barry:
As per the documentation string:
```
An INFO log is given when a container has a matching image:tag
An ERROR log is generated when a container does not match the image:tag
```
This is not what is happening. Currently, it only prints the `ERROR` log. This pull request updates the `kubectl image` command to also print an `INFO` log when the `image:tag` matches.

### DIFF
--- a/cmd/image.go
+++ b/cmd/image.go
@@ -17,12 +17,22 @@ type imgFlags struct {
 
 func printResultImg(results []Result) {
 	for _, result := range results {
-		if result.err > 0 {
+		switch result.err {
+		case 0:
 			log.WithFields(log.Fields{
 				"type": result.kubeType,
-				"tag":  result.img_tag}).Error(result.namespace,
+				"tag":  result.img_tag}).Info(result.namespace,
 				"/", result.name)
-
+		case 1:
+			log.WithFields(log.Fields{
+				"type": result.kubeType,
+				"tag":  result.img_tag}).Error("Image tag was missing ", result.namespace,
+				"/", result.name)
+		case 2:
+			log.WithFields(log.Fields{
+				"type": result.kubeType,
+				"tag":  result.img_tag}).Error("Image tag was incorrect ", result.namespace,
+				"/", result.name)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/image.go
```diff
@@ -92,6 +101,11 @@ Example usage:
 kubeaudit image --image gcr.io/google_containers/echoserver:1.7
 kubeaudit image -i gcr.io/google_containers/echoserver:1.7`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if len(imgConfig.img) == 0 {
+			log.Error("Empty image name. Are you missing the image flag?")
+			return
+		}
+
```
marc-barry: Without the above the error message was `Image tag is missing!` multiple times. This was confusing. In the above we check if the user specified `image:tag` with the `-i` or `--image` flags. It is also possible that the user uses the flag but passes an empty string.

cmd/image.go
```diff
@@ -15,14 +15,24 @@ type imgFlags struct {
...
 	for _, result := range results {
 		if result.err > 0 {
-			log.WithFields(log.Fields{
-				"type": result.kubeType,
-				"tag":  result.img_tag}).Error(result.namespace,
-				"/", result.name)
-
+			switch {
+			case result.img_tag != compImgTag:
```
marc-barry: `ERROR` if they don't match.

cmd/image.go
```diff
@@ -15,14 +15,24 @@ type imgFlags struct {
...
+				log.WithFields(log.Fields{
+					"type": result.kubeType,
+					"tag":  result.img_tag}).Error(result.namespace,
+					"/", result.name)
+			case result.img_tag == compImgTag:
```
marc-barry: `INFO` if they do match.

marc-barry:
I seem to have caused some failing tests with this pull request. Note sure how or why. Investigating.

marc-barry:
> I seem to have caused some failing tests with this pull request. Note sure how or why. Investigating.

Tests have been fixed so that they expect the current output.

jinankjain:
Thanks @marc-barry since we do the comparison part in `checkImage` function it would be nice to update the error code 1-> Image tag missing and 2-> Wrong Image tag. And then a switch case in `printResultImg` function. Why this:

- Firstly we don't need to pass `string image` to the function `printResultImg`
- Secondly, all the print function interfaces would be consistent and logic stays in those check functions.

jinankjain:
@klautcomputing @marc-barry I updated the PR with some changes.

marc-barry:
Thank you @jinankjain and @klautcomputing for taking care of the changes and shipping this. I've been unavailable this week.

marc-barry:
I don't think this pull request that was merged as is working as expected. I'll do another pull request with changes.

marc-barry:
> I don't think this pull request that was merged as is working as expected. I'll do another pull request with changes.

See #20.

PR #15, opened 2017-10-01
new feature: audit manifest file from command line args

jinankjain:
I started working on the new feature to audit config files passed as a command line argument. Right now it works only for securityContext. It needs to be ported to other features too.

I am not happy about of lot of stuff in this PR mostly regarding code quality. Need lot of interface love.  Sorry I was in hurry could not make code more clear.

It is just a proof of concept that is possible to do so.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -15,6 +15,7 @@ type rootFlags struct {
 	verbose    bool
 	allPods    bool
 	json       bool
+	manifest   string
 }
 
 var RootCmd = &cobra.Command{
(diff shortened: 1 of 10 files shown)


cmd/root.go
```diff
@@ -15,6 +15,7 @@ type rootFlags struct {
 	verbose    bool
 	allPods    bool
 	json       bool
+	config     string
```
JonPulsifer: I think it should be named something other than config, given that the names of the YAML files we anticipate to audit with this are called manifests.

cmd/root.go
```diff
@@ -41,4 +42,5 @@ func init() {
 	RootCmd.PersistentFlags().BoolVarP(&rootConfig.verbose, "verbose", "v", false, "Enable debug (verbose) logging")
 	RootCmd.PersistentFlags().BoolVarP(&rootConfig.json, "json", "j", false, "Enable json logging")
 	RootCmd.PersistentFlags().BoolVarP(&rootConfig.allPods, "allPods", "a", false, "Audit againsts pods in all the phases (default Running Phase)")
+	RootCmd.PersistentFlags().StringVarP(&rootConfig.config, "config", "y", "", "yaml configuration to audit")
```
JonPulsifer: I think the flag should be `-f`, staying in line w/ `kubectl` and other tools

klautcomputing:
@jinankjain Want to take a look and tell me what you think about my changes?

jinankjain:
Everything works now could be squashed into one commit and mergerd 👍  @klautcomputing

PR #19, opened 2017-10-05
New feature: audit yaml files with more than 1 resource

klautcomputing:
#15 Allowed us to audit maifests, unfortunately it only allowed for one resource to be present in the `manifest.yaml`

In this PR the functionality to audit manifest files that contain more than one resource is introduced.

### DIFF
--- a/fakeaudit/fakeDeployments.go
+++ b/fakeaudit/fakeDeployments.go
@@ -7,47 +7,31 @@ import (
 var deploymentsPath = filepath.Join(absPath, "fakeaudit", "test", "deployments")
 
 func CreateFakeDeploymentSC(namespace string) {
-	fakeDeploymentClient := getFakeDeploymentClient(namespace)
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC1.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC2.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC3.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC4.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC5.yml")))
+	yamls := []string{"fakeDeploymentSC1.yml", "fakeDeploymentSC2.yml", "fakeDeploymentSC3.yml", "fakeDeploymentSC4.yml", "fakeDeploymentSC5.yml"}
+	createHelper(namespace, deploymentsPath, yamls)
 }
 
 func CreateFakeDeploymentRunAsNonRoot(namespace string) {
-	fakeDeploymentClient := getFakeDeploymentClient(namespace)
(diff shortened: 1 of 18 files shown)


fakeaudit/utils.go
```diff
@@ -14,51 +15,97 @@ import (
 
 var absPath, _ = filepath.Abs("../")
 
-func ReadConfigFiles(filename string) runtime.Object {
+func ReadConfigFiles(filename string) (decoded []runtime.Object, err error) {
 	buf, err := ioutil.ReadFile(filename)
 
 	if err != nil {
 		log.Error("File not found")
 		panic(err)
```
jinankjain: Since there is an output err parameter we don't need to `panic` here just return would be good enough as it would get handled subsequently. 

Or we handle all the errors in this function and remove the output err parameter

fakeaudit/utils.go
```diff
@@ -14,51 +15,97 @@ import (
...
-	obj := ReadConfigFiles(filename)
-	deployment = obj.(*v1beta1.Deployment)
+func getDeployment(filename string) (deployment *v1beta1.Deployment, err error) {
+	obj_slice, err := ReadConfigFiles(filename)
+	if err != nil {
+		return
+	}
+	for _, obj := range obj_slice {
+		switch resource := obj.(type) {
+		case *v1beta1.Deployment:
```
jinankjain: Might also want to add a default to all these switch cases and add some sort of error like `Unsupported param`

jinankjain: Or since it's a single case why not if/else

jinankjain: Interesting I think all these functions like `getDeployment` etc can go away and we can just work with `getKubeResource` from cmd/utils.go and we can transfer that function in fakeaudit/utils.go

jinankjain: Yay! 🎉  lot of dead code removal

fakeaudit/fakeDeployments.go
```diff
@@ -4,50 +4,44 @@ import (
...
 
 func CreateFakeDeploymentSC(namespace string) {
-	fakeDeploymentClient := getFakeDeploymentClient(namespace)
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC1.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC2.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC3.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC4.yml")))
-	fakeDeploymentClient.Create(getDeployment(filepath.Join(deploymentsPath, "fakeDeploymentSC5.yml")))
+	yamls := []string{"fakeDeploymentSC2.yml", "fakeDeploymentSC1.yml", "fakeDeploymentSC3.yml", "fakeDeploymentSC4.yml", "fakeDeploymentSC5.yml"}
+	fakeDeploymentCreateHelper(namespace, yamls)
```
jinankjain: This refactoring is 😍

klautcomputing:
@jinankjain ready for a last review, I worked in the changes you suggested, I refactored a lot of stuff away and unified the error handling.

jinankjain:
Just one last thing squash everything into a single commit and ready to push. This PR was one of the most awaited PR.

klautcomputing:
Done which dismissed your review

PR #20, opened 2017-10-06
Print info for images by appending err==0 to the results

marc-barry:
In #13 this was supposed to fix what was shown in the [description](https://github.com/Shopify/kubeaudit/pull/13#issue-261690714) but it didn't do that. This pull request makes a small change to capture the `err==0` results and updates tests accordingly.

### DIFF
--- a/cmd/image.go
+++ b/cmd/image.go
@@ -66,11 +66,15 @@ func checkImage(container apiv1.Container, image string, result *Result) {
 
 	contTag = contImgLabel[1]
 
+	result.img_name = contImg
+	result.img_tag = contTag
+
 	if contImg == compImg && contTag != compTag {
 		result.err = 2
-		result.img_name = contImg
-		result.img_tag = contTag
 	}
+
+	result.err = 0
+
 	return
 }
 
@@ -79,7 +83,7 @@ func auditImages(image string, items Items) (results []Result) {
 		containers, result := containerIter(item)
 		for _, container := range containers {
 			checkImage(container, image, result)
-			if result != nil && result.err > 0 {
+			if result != nil && result.err >= 0 {
 				results = append(results, *result)
 				break
 			}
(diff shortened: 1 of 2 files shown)


cmd/image.go
```diff
@@ -79,7 +83,7 @@ func auditImages(image string, items Items) (results []Result) {
 		containers, result := containerIter(item)
 		for _, container := range containers {
 			checkImage(container, image, result)
-			if result != nil && result.err > 0 {
+			if result != nil && result.err >= 0 {
```
marc-barry: In order to get the `INFO` messages we need to look for `err` **`>=`** `0`.

cmd/image.go
```diff
@@ -66,11 +66,15 @@ func checkImage(container apiv1.Container, image string, result *Result) {
...
+	result.img_name = contImg
+	result.img_tag = contTag
+
 	if contImg == compImg && contTag != compTag {
 		result.err = 2
-		result.img_name = contImg
-		result.img_tag = contTag
 	}
+
+	result.err = 0
```
marc-barry: I know this is redundant since the zero value for an `int` is `0`. But I think it is better style to be explicit.

cmd/image.go
jinankjain: I am not sure about this since in the past we have only printed error messages i.e. faulty configurations. But I agree it is debatable. @klautcomputing

klautcomputing: Well this kind of doesn't make sense... `result.err` will always be `>= 0`(well should be, even though  in the code it's an `int` we should treat it as a `unit` and maybe change the code) so the check is redundant. Which means it was there for a reason because we only want to log errors.

But maybe we should either:
- have 2 different `log` methods? or
- have a `result.info` because this is not an error case.
if we want to allow non errors to be printed as well.

This question here is a bit hypothetical because no one but you uses the `image` programmatically but still I think we should come up with a way of handling more than the error case inside kubeaudit and find a real solution for this.

jinankjain: I have a way out for this one. Following more Linux style error handling like a big enum of errors that we support and every function should use from it.  Lots of wins for us by doing that:

- Centralized error tracking
- Adding more semantics to error rather than some bunch of number

jinankjain: Regarding that INFO that we should have a flag and then we will print INFO in all the cases. So that we have a unified interface everywhere

marc-barry:
Closing this pull request as this was covered in #26.

PR #25, opened 2017-10-14
Fixes #24: Replace auditSecurityContext with correct function calls

### LINKED ISSUE #24: Bug: auditing yml does not work
There is a bug in auditing yml as auditSecurityContext is invoked everywhere in image.go, runAsNonRoot.go etc instead of specific function

jinankjain:


klautcomputing:
Copy pasta 🍝  Thank you

PR #26, opened 2017-10-16
Refactor image

klautcomputing:
This is the first step towards error handling. Which lead me into a bit of a refactoring trip field trip.
Anyways, this enables us to easily integrate the functionality that @marc-barry wants to introduce with https://github.com/Shopify/kubeaudit/pull/20 but in a nicer way. I will update the PR once this one is merged.

This removes some splitting code with wasn't nice and `os.Exit(1)` from a function where it's way too late to do an `os.Exit` for a cmd line param check.

- [x] fix the commit message before merge to master
- [ ] squash commits

### DIFF
--- a/cmd/image.go
+++ b/cmd/image.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"os"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


cmd/image.go
```diff
@@ -37,44 +49,28 @@ func printResultImg(results []Result) {
...
 	}
 
-	contTag = contImgLabel[1]
-
-	if contImg == compImg && contTag != compTag {
+	if contImage.name == image.name && contImage.tag != image.tag {
 		result.err = 2
-		result.img_name = contImg
-		result.img_tag = contTag
+		result.img_name = contImage.name
```
jinankjain: imgName `go` likes this format

cmd/image.go
```diff
@@ -37,44 +49,28 @@ func printResultImg(results []Result) {
...
+		result.img_tag = contImage.tag
```
jinankjain: imgTag

jinankjain:
For more context on the above comment https://goreportcard.com/report/github.com/Shopify/kubeaudit

cmd/image.go
klautcomputing: That's not my code! I mean thanks for catching that :)

cmd/image.go
klautcomputing: `:GoRename to the rescue`

klautcomputing:
I just squashed the commits.

PR #29, opened 2017-10-16
Refactor runAsNonRoot to have named errors and better error messages

klautcomputing:
Same as #27 and #26

### DIFF
--- a/cmd/runAsNonRoot.go
+++ b/cmd/runAsNonRoot.go
@@ -8,24 +8,41 @@ import (
 
 func printResultNR(results []Result) {
 	for _, result := range results {
-		if result.err > 0 {
-			log.WithField("type", result.kubeType).Error(result.namespace, "/", result.name)
+		switch result.err {
+		case ErrorSecurityContextNIL:
+			log.WithFields(log.Fields{
+				"type":      result.kubeType,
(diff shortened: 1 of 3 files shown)


cmd/runAsNonRoot.go
```diff
@@ -8,23 +8,36 @@ import (
...
+	if container.SecurityContext.RunAsNonRoot == nil {
+		result.err = ErrorRunAsNonRootNIL
+		return
+	}
+	if !*container.SecurityContext.RunAsNonRoot {
+		result.err = ErrorRunAsNonRootFalse
+		return
 	}
-
 	return
```
csgregorian: ~return~

cmd/image.go
```diff
@@ -63,7 +63,7 @@ func checkImage(container apiv1.Container, image imgFlags, result *Result) {
 	}
 
 	if contImage.name == image.name && contImage.tag != image.tag {
-		result.err = 2
```
jinankjain: This should not be here!

klautcomputing: You mean in this PR? You are right it should have been in #26 ...

jinankjain: yeah or it should be moved to different commit

PR #31, opened 2017-10-16
Refactor automountServiceAccountToken

csgregorian:
Refactor automountServiceAccountToken audit to use proper error messages instead of numbers.

### DIFF
--- a/cmd/automountServiceAccountToken.go
+++ b/cmd/automountServiceAccountToken.go
@@ -5,30 +5,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func checkAutomountServiceAccountToken(result *Result) {
-
-	// Check for use of deprecated service account name
-	if result.dsa != "" {
-		result.err = 1
-	}
-
-	if result.token != nil {
-		// automountServiceAccountToken = true, and serviceAccountName is blank (default: default)
-		if *result.token && result.sa == "" {
-			result.err = 2
-		}
-	} else {
-		// automountServiceAccountToken = nil (default: true), and serviceAccountName is blank (default: default)
-		if result.sa == "" {
-			result.err = 3
-		}
-	}
-}
-
 func printResultASAT(results []Result) {
-
 	for _, result := range results {
-		if result.dsa != "" {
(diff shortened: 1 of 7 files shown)


cmd/automountServiceAccountToken.go
```diff
@@ -5,53 +5,50 @@ import (
...
+func checkAutomountServiceAccountToken(result *Result) {
+	// Check for use of deprecated service account name
+	if result.dsa != "" {
+		result.err = ErrorServiceAccountTokenDeprecated
+		return
+	}
 
+	if result.token != nil && *result.token && result.sa == "" {
+		// automountServiceAccountToken = true, and serviceAccountName is blank (default: default)
+		result.err = ErrorServiceAccountTokenNoName
```
jinankjain: Error name does not contain all the information. Need to cover both aspect of `and`

cmd/automountServiceAccountToken.go
```diff
@@ -5,53 +5,50 @@ import (
...
+		return
+	}
+
+	if result.token == nil && result.sa == "" {
+		// automountServiceAccountToken = nil (default: true), and serviceAccountName is blank (default: default)
+		result.err = ErrorServiceAccountTokenNIL
```
jinankjain: Same as previous one

cmd/automountServiceAccountToken_test.go
```diff
@@ -28,11 +28,11 @@ func TestDeploymentASAT(t *testing.T) {
 	}
 
 	for _, result := range results {
-		if result.name == "fakeDeploymentASAT1" && (result.err != 3 || result.dsa == "") {
+		if result.name == "fakeDeploymentASAT1" && (result.err != ErrorServiceAccountTokenDeprecated || result.dsa == "") {
```
jinankjain: This result.dsa == "" could go away and instead you can add another test for result.sa == ""

jinankjain:
@csgregorian You forgot to add the test for ASAT4 in autom.._test.go. ASAT4 is missing for rest of resources like DaemonSet etc.

PR #32, opened 2017-10-16
Refactored security context tests

csgregorian:
Refactored security context tests to replace numbered errors with constants.

### DIFF
--- a/cmd/errors.go
+++ b/cmd/errors.go
@@ -9,4 +9,6 @@ const (
 	ErrorReadOnlyRootFilesystemFalse
 	ErrorRunAsNonRootNIL
 	ErrorRunAsNonRootFalse
+	ErrorCapabilitiesNIL
+	ErrorCapabilitiesAddedOrNotDropped
 )
--- a/cmd/securityContext.go
+++ b/cmd/securityContext.go
@@ -12,52 +12,59 @@ func checkSecurityContext(container apiv1.Container, result *Result) {
 	result.capsDropped = true
 
 	if container.SecurityContext == nil {
-		result.err = 1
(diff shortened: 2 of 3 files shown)


cmd/securityContext.go
```diff
@@ -12,39 +12,37 @@ func checkSecurityContext(container apiv1.Container, result *Result) {
...
 	}
-
-	return
 }
 
 func printResultSC(results []Result) {
 	for _, result := range results {
 		switch err := result.err; err {
-		case 1:
+		case ErrorSecurityContextNIL:
```
klautcomputing: Can you standardize the logging to be in line with what we do in here: https://github.com/Shopify/kubeaudit/blob/master/cmd/image.go#L33

cmd/errors.go
```diff
@@ -9,4 +9,6 @@ const (
 	ErrorReadOnlyRootFilesystemFalse
 	ErrorRunAsNonRootNIL
 	ErrorRunAsNonRootFalse
+	ErrorCapabilitiesNIL
+	ErrorCapabilitiesAddedOrNotDropped
```
jinankjain: Not sure about this that we want to merge this two error condition or not? I remember having a discussion with @klautcomputing

klautcomputing: @csgregorian and I discussed that it's easier for now to just have one, get all the errors named in kubeaudit and then once this is done we will search for a way to support multiple errors in one result. That change could go hand in hand with the log level change and we could come up with a something that supports both multiple errors and different log levels.
But let's see about that.
I am fine to merge this as is and address it later.

jinankjain: cool

PR #36, opened 2017-10-17
Log levels

klautcomputing:
This introduces the concept of log levels, deletes a lot of duplicate tests, refactors a lot of code and introduces 

```
-v or --verbose to set the log level
--verbose=ERROR
--verbose=WARNING
--verbose=INFO
--verbose=DEBUG
```

The default log level is INFO and it will print INFO, WARNING and ERROR

Supported log levels are `info`, `warn` and `error` (for now, but more can be introduced easily if needed)

@marc-barry to confirm that what was meant to work in #13 actually works
@JonPulsifer to make sure all the features do what they should do (especially privilege needs a look and potentially some work)
@jinankjain and @csgregorian for reviewing my code


cmd/util.go
```diff
@@ -82,17 +83,59 @@ type kubeAuditReplicationControllers struct {
...
-	name        string
-	capsAdded   []apiv1.Capability
-	imgName     string
-	capsDropped bool
-	kubeType    string
-	dsa         string
-	sa          string
-	token       *bool
-	imgTag      string
+	Err         int
```
jinankjain: Is there any specific reason that you want to make these variable public instead of private?

klautcomputing: Yeah, afaik reflect only works on public variables, so to make the meta programming part work (that prints the correct fields of the `results` array) they needed to be exported.

jinankjain:
Could you please also  remove the test yml for other resources?

klautcomputing:
I will do that in one of my next PRs when I introduce a test framework, unless you feel strongly about doing it in this PR.

PR #42, opened 2017-10-20
Fixes #41: Update all dependencies to the current master

### LINKED ISSUE #41: Update all dependencies
Kubeaudit currently has a couple of older dependencies. Would be great to make Kubeaudit run on the never versions of all the dependencies.

jinankjain:
Update all dependencies to current master

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,85 +3,105 @@
 
 [[projects]]
   name = "cloud.google.com/go"
-  packages = ["compute/metadata","internal"]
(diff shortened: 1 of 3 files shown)


csgregorian:
Does this make it go gettable by fixing the kubernetes version we pull?

jinankjain:
How does it depend on `kubernetes`  version because `kubernetes` is not a dependency?

csgregorian:
The `k8s` go package versions, rather

jinankjain:
I think it should work assuming that I did not break anything in CI :crossed_fingers:

Issue #35, opened 2017-10-17
Check that all capabilities are dropped


klautcomputing:
with #34 we now have a list of all the capabilities that are dropped, now we should establish a way of making sure all possible capabilities are dropped. #33 would give us the functionality to specify that a cap wasn't dropped and that's intentional.

PR #45, opened 2017-10-23
Closes: #35 check that all capabilities were dropped, inform about non-dropped caps

klautcomputing:
In this PR we introduce the functionality to inform users about which capabilities were not dropped. There is a config file which allows users to specify the capabilities they want to drop.

#35 for reference


cmd/config
```diff
@@ -0,0 +1 @@
+../config
```
klautcomputing: Why is there a symlink you might ask! The answer is we need the file in the tests and the executable and they are in different places and I didn't feel like writing a config for where to find the config file

Gopkg.lock
```diff
@@ -176,6 +182,12 @@
   version = "v1.0.0"
 
 [[projects]]
+  name = "github.com/stretchr/testify"
```
klautcomputing: Does it really make sense to pull in `testify` when it's only used here once?! Yes, because we want to rewrite all the test!

cmd/securityContext.go
```diff
@@ -1,10 +1,52 @@
...
+		return
+	}
+	for _, toBeDroppedCap := range toBeDropped {
+		found := false
+		for _, droppedCap := range dropped {
+			if toBeDroppedCap == droppedCap {
+				found = true
+			}
+		}
+		if found == false {
```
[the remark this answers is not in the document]
klautcomputing: Nope I am just old-fashioned :)

PR #48, opened 2017-10-29
Refactored code, added getKubeResources() in util

natalysheinin:
- all occurrences of getKubeResources() renamed to getKubeResourcesManifest()
- new func getKubeResources() created in util.go to reduce code redundancy
- general code refactoring

### DIFF
--- a/cmd/image.go
+++ b/cmd/image.go
@@ -73,12 +73,9 @@ var imageCmd = &cobra.Command{
 	Use:   "image",
 	Short: "Audit container images",
 	Long: `This command audits a container against a given image:tag
-
 An INFO log is given when a container has a matching image:tag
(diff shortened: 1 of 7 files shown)


cmd/image.go
```diff
@@ -96,40 +93,28 @@ kubeaudit image -i gcr.io/google_containers/echoserver:1.7`,
 		if rootConfig.json {
 			log.SetFormatter(&log.JSONFormatter{})
 		}
+		var resources []Items
```
jinankjain: This whole thing could be wrapped in single function which would return a list of Items i.e. check if manifest is nil or not. Thus reducing all the redundant code.

klautcomputing: Yes, that's the next step :)

PR #53, opened 2017-10-30
Rewrite all the tests

### LINKED ISSUE #22: Refactoring of tests
Once #19 is merged we could get rid of all `fakeaudit/fakeResource.go` files with one helper function in `utils.go` that just traverses the `test` folder and builds very thing it needs on the fly. What do you think about this @jinankjain

klautcomputing:
Foreword: There is approximately 150 lines of code that need 👀 the rest is just `0xDEADC0DE` and fixing `yml` formatting.

This changes how we do tests for commands in `kubeaudit` from the ground up. Instead of relying on a complicated testing framework (`fakeaudit`) we now just read `manifest.yml`s and run the `auditX` functions on the result and make sure they create the correct `errors`.
This removes a lot of (now dead) code, unifies all the yamls from `fakeaudit/test` into `fixtures`, makes all test code smaller and more useful, also fixes tests for `image` which were broken.

this is ready for reviews :) let me know what you think, there is a lot of code that you don't need to review because it's now gone. Just give me a quick 👍 👎 for the new approach to how we do tests.

for reference this closes #22


cmd/automountServiceAccountToken_test.go
```diff
@@ -1,31 +1,15 @@
...
-	if len(results) != 2 {
-		t.Error("Test 1: Failed to detect all bad configuarations")
-	}
-
-	for _, result := range results {
-		if result.Name == "fakeReplicationControllerASAT1" && result.Occurrences[0].id != ErrorServiceAccountTokenDeprecated {
-			t.Error("Test 2: Failed to identify deprecated service account name. Refer: fakeReplicationControllerASAT1.yml")
-		}
+func TestServiceAccountTokenTrueAndNoName(t *testing.T) {
+	runTest(t, "../fixtures/service_account_token_true_and_no_name.yml", auditAutomountServiceAccountToken, ErrorServiceAccountTokenTrueAndNoName)
```
jinankjain: nit: paths package to join the path

Issue #52, opened 2017-10-30
Find a cleaner solution for the wait groups


klautcomputing:
Having the waitgroup code inside the function and outside the function is really ugly.
I am talking about this:
https://github.com/Shopify/kubeaudit/blob/master/cmd/runAsNonRoot.go#L39
and then
https://github.com/Shopify/kubeaudit/blob/master/cmd/runAsNonRoot.go#L76-L81

It would be nice if we could find a better solution for that.

PR #54, opened 2017-10-31
Refactor, clear up WaitGroup usage

josedonizetti:
Fixes https://github.com/Shopify/kubeaudit/issues/52. I've refactored it to in a way that I think is cleaner because:

- The WaitGroup usage is all in one place, one can see all the calls instead of having it on diff functions
- No global WaitGroup

Let me know what you think.

### DIFF
--- a/cmd/automountServiceAccountToken.go
+++ b/cmd/automountServiceAccountToken.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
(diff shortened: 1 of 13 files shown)


klautcomputing:
@josedonizetti Thanks a lot for doing this, this is way cleaner and exactly what I would have done 👍

PR #57, opened 2017-11-01
Refactor audit commands

### LINKED ISSUE #50: Refactor all commands
After #48, we can now refactor the commands even further take a look at: https://github.com/Shopify/kubeaudit/blob/master/cmd/runAsNonRoot.go#L56-L81
the same code will show up in basically all the commands and could be refactored into one function in `cmd/util.go` which additionally takes a function pointer as a parameter and then just calls that function in https://github.com/Shopify/kubeaudit/blob/master/cmd/runAsNonRoot.go#L79

fedorlis:
This closes #50 and #53

### DIFF
--- a/cmd/runAsNonRoot.go
+++ b/cmd/runAsNonRoot.go
@@ -1,9 +1,6 @@
 package cmd
 
 import (
-	"sync"
-
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/runAsNonRoot.go
```diff
@@ -77,13 +74,21 @@ kubeaudit runAsNonRoot`,
 		var wg sync.WaitGroup
 		wg.Add(len(resources))
 
+		results := make(chan Result)
```
jinankjain: Finally some use of go channels :heart_eyes:

PR #62, opened 2017-11-03
Update commands and update readme

klautcomputing:
let me know what you think, since this is mostly a non coding PR I would appreciate some feedback about whether you like the renaming of the commands/files and whether you like the updated readme. I still have to do some proof reading and flesh out some command descriptions inside the readme as well as the `Long` description inside the commands itself.


README.md
```diff
@@ -2,167 +2,218 @@
...
+  - [Audit readOnlyRootFilesystem](#rootfs)
+  - [Audit runAsNonRoot](#root)
+  - [Audit privileged](#priv)
+  - [Audit capabilities](#caps)
+- [Audit image](#image)
+- [Audit Service Accounts](#sat)
+- [Audit network policies](#netpol)
 
 <a name="sc" />
+#### Audit security contexts
```
jinankjain: Extra \n

PR #68, opened 2017-11-06
Error when AllowPrivilegeEscalation is nil

fedorlis:
When `AllowPrivilegeEscalation` is nil privilege escalation is allowed, so we should error instead of warning.


klautcomputing:
Sorry for missing this :)

PR #71, opened 2017-11-16
Quota cmd

jerr:
Hi,

Setting CPU and memory limits is a good security practice. (see http://blog.kubernetes.io/2016/08/security-best-practices-kubernetes-deployment.html)
Why not test this with a kubeaudit command?

What do you think of this notion and my implementation?
Your feedbacks are very welcome.

Regards,
Jeremie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ audits against pods in all the phases (default Running Phase)
 - [Audit image](#image)
 - [Audit Service Accounts](#sat)
 - [Audit network policies](#netpol)
+- [Audit resources](#resources)
 
 <a name="sc" />
 
@@ -247,6 +248,27 @@ for more information:
 # don't specify -l or -c to run inside the clsuter
 kubeaudit np
 WARN[0000] Default allow mode on test/testing
+```
+
+<a name="resources" />
+
+### Audit resources limits
+
+It checks that every resource has a CPU and memory limit. See [Kubernetes Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/)
+for more information:
+
+```sh
+kubeaudit -l limits
+WARN[0000] CPU limit not set, please set it!
+WARN[0000] Memory limit not set, please set it!
+```
+
+With the `--cpu` and `--memory` parameters, `kubeaudit` can check the limits not to be exceeded.
+
+```sh
+kubeaudit -l limits --cpu 500m --memory 125Mi
+WARN[0000] CPU limit exceeded, it is set to 1 but it must not exceed 500m. Please adjust it! !
+WARN[0000] Memory limit exceeded, it is set to 512Mi but it must not exceed 125Mi. Please adjust it!
 ```
 
 <a name="contribute" />
--- a/cmd/limits.go
+++ b/cmd/limits.go
@@ -1,0 +1,120 @@
+package cmd
+
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	k8sResource "k8s.io/apimachinery/pkg/api/resource"
+)
+
+type limitFlags struct {
+	cpuArg    string
+	cpu       k8sResource.Quantity
+	memoryArg string
+	memory    k8sResource.Quantity
+}
+
+var limitConfig limitFlags
+
+func (limit *limitFlags) parseLimitFlags() {
+	if len(limit.cpuArg) != 0 {
+		quantity, err := k8sResource.ParseQuantity(limit.cpuArg)
+		if err != nil {
+			log.Error("Wrong cpu argument: " + err.Error())
+			return
+		}
+		limit.cpu = quantity
+	}
+
+	if len(limit.memoryArg) != 0 {
+		quantity, err := k8sResource.ParseQuantity(limit.memoryArg)
+		if err != nil {
+			log.Error("Wrong cpu argument: " + err.Error())
+			return
+		}
+		limit.memory = quantity
+	}
+}
+
+func checkLimits(container Container, limits limitFlags, result *Result) {
+	if container.Resources.Limits == nil {
+		occ := Occurrence{id: ErrorResourcesLimitsNIL, kind: Warn, message: "Resource limit not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	checkCPULimit(container, limits, result)
+	checkMemoryLimit(container, limits, result)
+}
+
+func checkCPULimit(container Container, limits limitFlags, result *Result) {
+	cpu := container.Resources.Limits.Cpu()
+	if cpu == nil || cpu.IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitsCpuNIL, kind: Warn, message: "CPU limit not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	if limits.cpu.MilliValue() > 0 && cpu.MilliValue() > limits.cpu.MilliValue() {
+		result.CPULimitActual = cpu.String()
+		result.CPULimitMax = limits.cpu.String()
+		message := fmt.Sprintf("CPU limit exceeded, it is set to %s but it must not exceed %s. Please adjust it!", cpu.String(), limits.cpu.String())
+		occ := Occurrence{id: ErrorResourcesLimitsCpuExceeded, kind: Warn, message: message}
+		result.Occurrences = append(result.Occurrences, occ)
+	}
+}
+
+func checkMemoryLimit(container Container, limits limitFlags, result *Result) {
+	memory := container.Resources.Limits.Memory()
+	if memory == nil || memory.IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitsMemoryNIL, kind: Warn, message: "Memory limit not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	if limits.memory.Value() > 0 && memory.Value() > limits.memory.Value() {
+		result.MEMLimitActual = memory.String()
+		result.MEMLimitMax = limits.memory.String()
+		message := fmt.Sprintf("Memory limit exceeded, it is set to %s but it must not exceed %s. Please adjust it!", memory.String(), limits.memory.String())
+		occ := Occurrence{id: ErrorResourcesLimitsMemoryExceeded, kind: Warn, message: message}
+		result.Occurrences = append(result.Occurrences, occ)
+	}
+}
+
+func auditLimits(limits limitFlags, items Items) (results []Result) {
+	limits.parseLimitFlags()
+
+	for _, item := range items.Iter() {
+		containers, result := containerIter(item)
+		for _, container := range containers {
+			checkLimits(container, limits, result)
+			if result != nil && len(result.Occurrences) > 0 {
+				results = append(results, *result)
+				break
+			}
+		}
+	}
+	return
+}
+
+var limitsCmd = &cobra.Command{
+	Use:   "limits",
+	Short: "Audit containers running with limits",
+	Long: `This command determines which containers in a kubernetes cluster have and do not exceed specified cpu and memory limits.
+
+A PASS is given when a container has cpu and memory limits
+A FAIL is given when a container does not have cpu and memory limits
+
+Example usage:
+kubeaudit limits
+kubeaudit limits --cpu 500m --memory 256Mi`,
+	Run: runAudit(auditLimits),
+}
+
+func init() {
+	RootCmd.AddCommand(limitsCmd)
+	limitsCmd.Flags().StringVar(&limitConfig.cpuArg, "cpu", "", "max cpu limit")
+	limitsCmd.Flags().StringVar(&limitConfig.memoryArg, "memory", "", "max memory limit")
+}
--- a/cmd/limits_test.go
+++ b/cmd/limits_test.go
@@ -1,0 +1,22 @@
+package cmd
+
+import "testing"
+
+func TestResourcesLimitsNil(t *testing.T) {
+	runTest(t, "resources_limit_nil.yml", auditLimits, ErrorResourcesLimitsNIL)
+}
+
+func TestResourcesNoCPULimit(t *testing.T) {
+	runTest(t, "resources_limit_no_cpu.yml", auditLimits, ErrorResourcesLimitsCpuNIL)
+}
+
+func TestResourcesNoMemoryLimit(t *testing.T) {
+	runTest(t, "resources_limit_no_memory.yml", auditLimits, ErrorResourcesLimitsMemoryNIL)
+}
+func TestResourcesCPULimitExceeded(t *testing.T) {
+	runTest(t, "resources_limit.yml", auditLimits, ErrorResourcesLimitsCpuExceeded, "600m", "")
+}
+
+func TestResourcesMemoryLimitExceeded(t *testing.T) {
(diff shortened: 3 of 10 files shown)


jinankjain:
Hey @jerr,

Thanks for the awesome PR!

It would be great if you could also add an option for CPU and memory bounds.

For ex:

`kubeaudit -l quota -cpu 10 -mem 256Mi` something like that. That would be a really awesome feature.

jerr:
Hi @jinankjain

I'll try to add options for max CPU ND memory limits. Thanks for your feedback.

I have also other ideas for this command. We can check quota and quotarange in the future. That's why I named the command "quota" not "limit".
For gpu and ephemeral storage limits in need to study a little more use cases.

Regards
Jérémie

README.md
```diff
@@ -249,6 +250,19 @@ kubeaudit np
 WARN[0000] Default allow mode on test/testing
 ```
 
+<a name="resources" />
+
+### Audit resource quotas
+
+It checks that every resources should have a CPU and memory limts. See [Kubernetes Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/)
```
klautcomputing: Typo, change to: `It checks that every resource has a CPU and memory limit.`

cmd/quota.go
```diff
@@ -0,0 +1,63 @@
...
+
+func checkQuota(container Container, result *Result) {
+	if container.Resources.Limits == nil {
+		occ := Occurrence{id: ErrorResourcesLimitNIL, kind: Error, message: "Quota not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	if container.Resources.Limits.Cpu() == nil || container.Resources.Limits.Cpu().IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitNIL, kind: Error, message: "CPU quota not set, please set it!"}
```
klautcomputing: I think we should have 3 different codes for this: `ResourcesLimitsNil`, `ResourcesLimitsCpuNIL` and `ResourcesLimitsMemoryNIL`

cmd/quota.go
```diff
@@ -0,0 +1,63 @@
...
+		result.Occurrences = append(result.Occurrences, occ)
+	}
+
+	if container.Resources.Limits.Memory() == nil || container.Resources.Limits.Memory().IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitNIL, kind: Error, message: "Memory quota not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+	}
+
+	if len(result.Occurrences) == 0 {
```
klautcomputing: I am not sure we want to log an info message if everything is configured correctly because we don't do that for any other command. The only time we log is for an image audit and there only because it makes sense to tell that the matching image was found.

README.md
```diff
@@ -249,6 +250,19 @@ kubeaudit np
...
+for more information:
+
+```sh
+kubeaudit -l quota
+ERRO[0000] CPU quota not set, please set it!
```
klautcomputing: I think these should be `WARN`ings and not `ERRO`rs. I am pulling in @JonPulsifer for some feedback on this

cmd/quota_test.go
```diff
@@ -0,0 +1,15 @@
+package cmd
+
+import "testing"
+
+func TestQuotaNil(t *testing.T) {
```
klautcomputing: I am just leaving a note here that if you split the status into 3 statuses we would also need more tests :)

README.md
jerr: Maybe the Warn level is better in this case, I will change it.

cmd/quota.go
jerr: Ok for 3 different codes, plus two other codes for the parameters case. ResourcesLimitsCpuExceeded and ResourcesLimitsMemoryExceeded.

cmd/quota.go
jerr: I add to test when cpu and memory limits are ok (like image command). But I prefer to find a better way. I remove it.

jerr:
Thank you @klautcomputing  and @jinankjain,

I'm applying your feedback, I still need to improve the code and the tests. I hope to finish this weekend.
Just a quick question, if I use the warning level, should the name of the const start with Warn? like WarnResourcesLimitsCpuExceeded or Error as usual like ErrorResourcesLimitsCpuExceeded? 

Jeremie.

klautcomputing:
> Just a quick question, if I use the warning level, should the name of the const start with Warn?

I don't care :) I mean it makes sense to use `WarnXYZ`, but then again I didn't do it... So feel free to go either way unless someone else expresses a strong opinion about it.

jerr:
Hope that better with CPU and memory options for quota command.

Regards,
Jeremie.

cmd/test_util.go
```diff
@@ -10,17 +10,24 @@ import (
...
-		if len(imageStr) != 1 {
+		if len(argStr) != 1 {
 			log.Fatal("Incorrect number of images specified")
 		}
-		image = imgFlags{img: imageStr[0]}
+		image = imgFlags{img: argStr[0]}
 		image.splitImageString()
+	case (func(quotaFlags, Items) []Result):
+		if len(argStr) != 2 {
+			log.Fatal("Incorrect number of images specified")
```
klautcomputing: You forgot to change the string :)

cmd/quota.go
```diff
@@ -0,0 +1,116 @@
...
+
+Example usage:
+kubeaudit quota
+kubeaudit quota --cpu 500m --memory 256Mi`,
+	Run: runAudit(auditQuota),
+}
+
+func init() {
+	RootCmd.AddCommand(quotaCmd)
+	quotaCmd.Flags().StringVar(&quotaConfig.cpuArg, "cpu", "", "max cpu limit")
```
klautcomputing: Could you move them into the root command: https://github.com/Shopify/kubeaudit/blob/master/cmd/root.go as well.

cmd/quota.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	k8sResource "k8s.io/apimachinery/pkg/api/resource"
+)
+
+var quotaConfig quotaFlags
+
+type quotaFlags struct {
```
klautcomputing: Let's move the variables into https://github.com/Shopify/kubeaudit/blob/master/cmd/root.go we decided to keep all the variables there. And did so for all the other commands.

klautcomputing:
One last thing: Would you do me the favor and squash all your commits into one? :)

jerr:
I made the changes, tell me if it fits your expectations, especially for the root.go file.

Regards,
Jeremie.

cmd/quota.go
jinankjain: @klautcomputing but cpu and memory limits are not root flags like other flags in root.go. They are bind to quota command. So I am not sure about moving cpu and mem flags to root.go?

jerr: I'm not sure for that too. I can reintegrate it into quota command if it's necessary. Let me know.

cmd/quota.go
```diff
@@ -0,0 +1,108 @@
...
+	if quota.cpu.MilliValue() > 0 && cpu.MilliValue() > quota.cpu.MilliValue() {
+		occ := Occurrence{id: ErrorResourcesLimitsCpuExceeded, kind: Warn, message: "CPU limit exceeded, please adjust it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+	}
+}
+
+func checkMemoryLimit(container Container, quota quotaFlags, result *Result) {
+	memory := container.Resources.Limits.Memory()
+	if memory == nil || memory.IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitsMemoryNIL, kind: Warn, message: "Memory quota not set, please set it!"}
```
klautcomputing: Why is this called `quota` here? Because what we seem to check from the code is `container.Resources.Limits.Memory()`

cmd/quota.go
```diff
@@ -0,0 +1,108 @@
...
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	if quota.memory.Value() > 0 && memory.Value() > quota.memory.Value() {
+		occ := Occurrence{id: ErrorResourcesLimitsMemoryExceeded, kind: Warn, message: "Memory limit exceeded, please adjust it!"}
```
klautcomputing: Here it seems to be exactly the other way round, the code checks for `quota.memory.Value()` and then logs about `Memory limits`.

karanthukral:
I am trying to understand what the motivation behind the addition is. I think it is confusing `ResourceQuotas` and `Resource Requests` and `Resource Limits`. `ResorceQuotas` are quotas that are applied as sum for all running pods in a namespace. All pods together can't invalidate the quota. Here it seems the the tool is just trying to check if the resource limits are set in a template for a container and then comparing it against passed in values. 

Would someone be able to clarify more the motivation for this?

cmd/quota.go
jerr: I'll add more information in the message and also see why we have duplicates in your case. Let me come back to that.

jerr:
@karanthukral my point of view on the command name:
For me, **Quota** is the generic term to describe the concept of resource limitation. This includes the different parameters such as the notions of **limit**, **requested**, **quota** and **quota range**.

The question is: do we need a command or several commands to handle this.
I chose to use a single command and for the moment only the verification of the limit setting is checked.
What do you think of this choice?

karanthukral:
> What do you think of this choice?

I see what you mean. I would say in terms of groupings, requests and limits are valid to group since they live in a single template. Since resource quotas in k8s are for a namespace and not a single template, that should be handled separately.

What do you think about using `resources` as the command instead of `quota`?

jerr:
@karanthukral yes, `resources` maybe it's better.
If I understand you correctly, a command name should relate to the item to be checked and not what it does. Like `image` or `caps` commands.

I'll try with `resources` (I could go back to the `quota` or something else if needed).

Naming is very important, so no problem if this PR is not accepted right away.

jerr:
I made some changes. 
- rename the command to `resources`
- add additional information in messages

Maybe the command name `resources` is too general, `limits` would be more appropriate.

README.md
```diff
@@ -249,6 +250,27 @@ kubeaudit np
 WARN[0000] Default allow mode on test/testing
 ```
 
+<a name="resources" />
+
+### Audit resources
+
+It checks that every resource has a CPU and memory limit. See [Kubernetes Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/)
```
karanthukral: Maybe its better to link [this](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/) instead of the resourcequota docs

karanthukral:
> Maybe the command name resources is too general, limits would be more appropriate.

The reason `resources` could work is if we have sub commands to check `limit` and `requests`. If not `resource-limits` or `limits` is a valid command name. 

I am not to familiar with the codebase, but the idea seems valid to me and the rename of the command to not mention `ResourceQuotas` helps a lot. Thanks :D

karanthukral:
Another question I had was what is the idea behind having arbitrary defined bounds for memory and cpu limits? Is it simply to run through all templates and make sure no one goes over some numbers?

jerr:
Hi all,

I renamed the command to "limits". It's simpler and corresponds to what it does.
For other aspects of quotas, we can do that in other commands.

@klautcomputing,  I left `limitConfig` variable and `limitFlags` structure in `limits.go` file because they are specific to this command. I hope that suits you.


_**Here is a simple use case of the command:**_ 
`./kubeaudit -l limits --cpu 100m --memory 128Mi`
```
WARN[0000] CPU limit exceeded, it is set to 200m but it must not exceed 100m. Please adjust it!  KubeType=replicationController Name=docker-registry-2 Namespace=default
WARN[0000] Memory limit exceeded, it is set to 512Mi but it must not exceed 128Mi. Please adjust it!  KubeType=replicationController Name=docker-registry-2 Namespace=default
WARN[0000] Resource limit not set, please set it!        KubeType=replicationController Name=router-1 Namespace=default
```

Regards,
Jeremie.

cmd/limits_test.go
```diff
@@ -0,0 +1,22 @@
+package cmd
+
+import "testing"
+
+func TestResourcesLimitsNil(t *testing.T) {
+	runTest(t, "resources_limit_nil.yml", auditLimits, ErrorResourcesLimitsNIL, "", "")
```
klautcomputing: `runTest` has optional parameters so the last arguments `, "", ""` are not needed.

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+	}
+	return
+}
+
+var limitsCmd = &cobra.Command{
+	Use:   "limits",
+	Short: "Audit containers running with limits",
+	Long: `This command determines which containers in a kubernetes cluster
```
fedorlis: 👋 Hey thanks for putting up an awesome PR, just wanted to add that the description should probably be something closer to `This command determines which containers in a kubernetes cluster have and do not exceed specified cpu and memory limits`. The command does more than just verify the limits' existence.

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+	if cpu == nil || cpu.IsZero() {
+		occ := Occurrence{id: ErrorResourcesLimitsCpuNIL, kind: Warn, message: "CPU limit not set, please set it!"}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+
+	if limits.cpu.MilliValue() > 0 && cpu.MilliValue() > limits.cpu.MilliValue() {
+		message := fmt.Sprintf("CPU limit exceeded, it is set to %s but it must not exceed %s. Please adjust it!", cpu.String(), limits.cpu.String())
+		occ := Occurrence{id: ErrorResourcesLimitsCpuExceeded, kind: Warn, message: message}
+		result.Occurrences = append(result.Occurrences, occ)
```
klautcomputing: If you want to make it easy for people to parse the violations could log them in the result object and them print them as for example CPULimit, CPUActual or something like this take a look at this: https://github.com/Shopify/kubeaudit/blob/master/cmd/util.go#L119-L148

cmd/test_util.go
```diff
@@ -10,17 +10,24 @@ import (
...
 	case (func(imgFlags, Items) []Result):
-		if len(imageStr) != 1 {
+		if len(argStr) != 1 {
 			log.Fatal("Incorrect number of images specified")
 		}
-		image = imgFlags{img: imageStr[0]}
+		image = imgFlags{img: argStr[0]}
 		image.splitImageString()
+	case (func(limitFlags, Items) []Result):
+		if len(argStr) != 2 {
```
fedorlis: The test doesn't always need limit parameters

cmd/limits.go
jerr: I didn't see that. It's better to process the results.

fixtures/resources_limit_no_memory.yml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        apps: fakeNoLimitQuota
+    spec:
+      containers:
+      - name: fakeContainerNoMemoryLimit
+        resources: 
+          requests:
```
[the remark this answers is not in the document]
jerr: I added this to have an example of requested resources but it does not have any interest in the test. I remove it.

cmd/limits.go
jerr: Thank you, I didn't update the description after adding the parameters.

cmd/test_util.go
jerr: For the arguments, it would be nice to be able to name them. But this is not the subject of this PR.
To be simple, the code accepts 0 or 2 arguments.

PR #72, opened 2017-12-04
Auditing for specific namespace (with -n or --namespace option)

jerr:
Limiting auditing to a namespace can be useful in some cases.
I propose to add the `-n` or `--namespace` option to do this.

```
kubeaudit -l sat -n default
ERRO[0000] Default serviceAccount with token mounted. Please set AutomountServiceAccountToken to false  KubeType=deployment Name=redis Namespace=default
```
If nothing is indicated, all namespaces are taken into account as currently.

What do you think about this option?

### DIFF
--- a/cmd/namespace_test.go
+++ b/cmd/namespace_test.go
@@ -1,0 +1,37 @@
+package cmd
+
+import (
+	"testing"
+)
+
+func TestDaemonSetInNamespace(t *testing.T) {
+	runTestInNamespace(t, "fakeDaemonSetPrivileged", "privileged_true.yml", auditPrivileged, ErrorPrivilegedTrue)
+}
+
+func TestDaemonSetNotInNamespace(t *testing.T) {
+	runTestInNamespace(t, "otherFakeDaemonSetPrivileged", "privileged_true.yml", auditPrivileged)
+}
+
+func TestDeploymentInNamespace(t *testing.T) {
+	runTestInNamespace(t, "fakeDeploymentSC", "capabilities_some_dropped.yml", auditCapabilities, ErrorCapabilitiesSomeDropped)
+}
+
+func TestDeploymentNotInNamespace(t *testing.T) {
+	runTestInNamespace(t, "otherFakeDeploymentSC", "capabilities_some_dropped.yml", auditCapabilities)
+}
+
+func TestStatefulSetInNamespace(t *testing.T) {
+	runTestInNamespace(t, "fakeStatefulSetRORF", "read_only_root_filesystem_nil.yml", auditReadOnlyRootFS, ErrorReadOnlyRootFilesystemNIL)
+}
+
+func TestStatefulSetNotInNamespace(t *testing.T) {
+	runTestInNamespace(t, "otherFakeStatefulSetRORF", "read_only_root_filesystem_nil.yml", auditReadOnlyRootFS)
+}
+
+func TestReplicationControllerInNamespace(t *testing.T) {
+	runTestInNamespace(t, "fakeReplicationControllerASAT", "service_account_token_nil_and_no_name.yml", auditAutomountServiceAccountToken, ErrorServiceAccountTokenNILAndNoName)
+}
+
+func TestReplicationControllerNotInNamespace(t *testing.T) {
+	runTestInNamespace(t, "otherFakeReplicationControllerASAT", "service_account_token_nil_and_no_name.yml", auditAutomountServiceAccountToken)
+}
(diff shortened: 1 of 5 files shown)


jinankjain:
Can you add some test for this in the test suite?

jerr:
Hi @jinankjain 
I'm adding tests to get better code coverage but it's not ready yet.
Let me come back to that.

jerr:
I added some namespace tests.
To put it simply, if the error code is negative then it must not appear in the audit list of detected errors.

Let me know if that suits you.

cmd/test_util.go
```diff
@@ -58,5 +59,10 @@ func runTest(t *testing.T, file string, function interface{}, errCode int, argSt
 			errors = append(errors, occurrence.id)
 		}
 	}
-	assert.Contains(errors, errCode)
+	if errCode > 0 {
+		assert.Contains(errors, errCode)
+	} else {
+		assert.NotContains(errors, -errCode)
```
jinankjain: Isn't error codes always positive as they are part of that pesudo enum?

cmd/util.go
```diff
@@ -22,44 +24,53 @@ func debugPrint() {
...
 }
 
+func isInNamespace(meta metav1.ObjectMeta) (valid bool) {
+	return rootConfig.namespace == apiv1.NamespaceAll || rootConfig.namespace == meta.Namespace
+}
+
 func convertDeploymentToDeploymentList(deployment Deployment) (deploymentList *DeploymentList) {
-	deploymentList = &DeploymentList{
-		Items: []Deployment{deployment},
+	if isInNamespace(deployment.ObjectMeta) {
```
jinankjain: I have a better suggestion for these `convertBla` should return types like `kubeAuditBla` instead of k8s types. This will help in simplifying the code. As later in the code we use convert function to construct those types. Other suggestion would be just get rid of those custom types like `kubeAuditBla` and only use native types from k8s, which would require a lot of refactoring but possible with one type like interface{}[].

jerr:
I changed the tests a bit of namespace_test.go for less impacted existing code (especially in test_util.go).
I hope it's simpler like that.

I'm thinking of improving the runTest function to be able to address more scenarios but it's better to do it in another pull request (like to validate the other data of the results).

cmd/util.go
jerr: I looked at your second suggestion and it seems possible and simplifies the code. Maybe you can create a refactoring issue for that.

cmd/namespace_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestReplicationControllerInNamespace(t *testing.T) {
+	runTestInNamespace(t, "fakeReplicationControllerASAT", "service_account_token_nil_and_no_name.yml", auditAutomountServiceAccountToken, ErrorServiceAccountTokenNILAndNoName)
+}
+
+func TestReplicationControllerNotInNamespace(t *testing.T) {
+	runTestInNamespace(t, "otherFakeReplicationControllerASAT", "service_account_token_nil_and_no_name.yml", auditAutomountServiceAccountToken)
+}
+
+func runTestInNamespace(t *testing.T, namespace string, file string, function interface{}, errCode ...int) {
```
klautcomputing: This should be in `test_util.go`

cmd/test_util.go
```diff
@@ -58,5 +58,9 @@ func runTest(t *testing.T, file string, function interface{}, errCode int, argSt
 			errors = append(errors, occurrence.id)
 		}
 	}
-	assert.Contains(errors, errCode)
+
+	if errCode != 0 {
```
klautcomputing: Good catch 👍  our errors starts at 1

cmd/util.go
```diff
@@ -22,44 +24,53 @@ func debugPrint() {
 	}
 }
 
+func isInNamespace(meta metav1.ObjectMeta) (valid bool) {
```
klautcomputing: Could we generalize this function so that it takes a namespace as an argument? Because from reading the function name I expected it to take a namespace as an argument.

And then maybe have a second function `isInRootConfigNamespace` (not the best name, but I guess you get the idea) that calls just `isInNamespace(rootConfig.namespace)`

TLDR & easy solution: rename the function :)

cmd/util.go
klautcomputing: I am in favor of merging this imperfect solution and address the underlying and deeper refactoring issue in another (and way way longer) PR.

cmd/util.go
jerr: I have no better proposal for the moment so I'll take yours.

klautcomputing:
I merged this this and we need to address the concern that @jinankjain brought up (that the code gets uglier) in a separate (set of) PR(s)

PR #73, opened 2017-12-19
Add options to override audits using labels

fedorlis:
This PR adds support for the following labels:
    - `kubeaudit.allow.privilegeEscalation`: Allows setting `AllowPrivilegeEscalation` to true
    - `kubeaudit.alllow.privileged`: Allows setting `Privileged` to true
    - `kubeaudit.allow.capability.{{capName}}`: Allows the addition/retention of specified capability
    - `kubeaudit.allow.runAsRoot`: Allows setting `runAsNonRoot` to false
    - `kubeaudit.allow.automountServiceAccountToken`: Allows setting `automountServiceAccountToken` to true
    - `kubeaudit.allow.readOnlyRootFilesystemFalse`: Allows setting readOnlyRootFilesystem to false  

The labels will override their respective audits when their value is not nil. The value will be interpreted as the `Reason` said override was implemented, with a value of `"True"` or `"true"` being interpreted as `"Unspecified"`.

### DIFF
--- a/cmd/allowPrivilegeEscalation.go
+++ b/cmd/allowPrivilegeEscalation.go
@@ -7,20 +7,48 @@ import (
 
 func checkAllowPrivilegeEscalation(container Container, result *Result) {
 	if container.SecurityContext == nil {
-		occ := Occurrence{id: ErrorSecurityContextNIL, kind: Error, message: "SecurityContext not set, please set it!"}
+		occ := Occurrence{
+			id:      ErrorSecurityContextNIL,
+			kind:    Error,
+			message: "SecurityContext not set, please set it!",
+		}
 		result.Occurrences = append(result.Occurrences, occ)
 		return
 	}
-	if container.SecurityContext.AllowPrivilegeEscalation == nil {
-		occ := Occurrence{id: ErrorAllowPrivilegeEscalationNIL, kind: Error, message: "AllowPrivilegeEscalation not set which allows privilege escalation, please set to false"}
+	if reason := result.Labels["kubeaudit.allow.privilegeEscalation"]; reason == "" {
+		if container.SecurityContext.AllowPrivilegeEscalation == nil {
+			occ := Occurrence{
+				id:      ErrorAllowPrivilegeEscalationNIL,
+				kind:    Error,
+				message: "AllowPrivilegeEscalation not set which allows privilege escalation, please set to false",
+			}
+			result.Occurrences = append(result.Occurrences, occ)
+		} else if *container.SecurityContext.AllowPrivilegeEscalation == true {
+			occ := Occurrence{
+				id:      ErrorAllowPrivilegeEscalationTrue,
+				kind:    Error,
+				message: "AllowPrivilegeEscalation set to true, please set to false",
+			}
+			result.Occurrences = append(result.Occurrences, occ)
+		}
+	} else if container.SecurityContext.AllowPrivilegeEscalation == nil || *container.SecurityContext.AllowPrivilegeEscalation == true {
+		occ := Occurrence{
+			id:       ErrorAllowPrivilegeEscalationTrueAllowed,
+			kind:     Warn,
+			message:  "Allowed AllowPrivilegeEscalation to be set as true",
+			metadata: Metadata{"Reason": prettifyReason(reason)},
+		}
 		result.Occurrences = append(result.Occurrences, occ)
-		return
-	}
-	if *container.SecurityContext.AllowPrivilegeEscalation {
-		occ := Occurrence{id: ErrorAllowPrivilegeEscalationTrue, kind: Error, message: "AllowPrivilegeEscalation set to true, please set to false"}
+	} else {
+		occ := Occurrence{
+			id:       ErrorMisconfiguredKubeauditAllow,
+			kind:     Warn,
+			message:  "Allowed setting AllowPrivilegeEscalation to true, but it is set to false",
+			metadata: Metadata{"Reason": prettifyReason(reason)},
+		}
 		result.Occurrences = append(result.Occurrences, occ)
-		return
 	}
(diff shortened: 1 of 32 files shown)


cmd/automountServiceAccountToken.go
```diff
@@ -12,16 +12,22 @@ func checkAutomountServiceAccountToken(result *Result) {
 		return
 	}
 
+	if result.Labels["kubeaudit.allow.automountServiceAccountTokenTrue"] != "" {
+		occ := Occurrence{id: ErrorAutomountServiceAccountTokenTrueAllowed, kind: Warn, message: "Allowed setting AutomountServiceAccountToken to true", metadata: map[string]string{"Reason": result.Labels["kubeaudit.allow.automountServiceAccountToken"]}}
```
jinankjain: nit: one line per field

cmd/privileged.go
```diff
@@ -6,18 +6,39 @@ import (
...
-		occ := Occurrence{id: ErrorSecurityContextNIL, kind: Error, message: "SecurityContext not set, please set it!"}
+		occ := Occurrence{
+			id:      ErrorSecurityContextNIL,
+			kind:    Error,
+			message: "SecurityContext not set, please set it!",
+		}
+		result.Occurrences = append(result.Occurrences, occ)
+		return
+	}
+	if reason := result.Labels["kubeaudit.allow.privileged"]; reason != "" {
```
klautcomputing: same as above, we don't beautify the reason, but check nicer

cmd/result.go
```diff
@@ -7,29 +7,27 @@ import (
...
+	CPULimitActual string
+	CPULimitMax    string
+	DSA            string
 	Err            int
-	Occurrences    []Occurrence
-	Namespace      string
-	Name           string
-	CapsAdded      []Capability
 	ImageName      string
-	CapsDropped    []Capability
```
klautcomputing: Where did the `Caps...` variables go? Do we now just use: `container.SecurityContext.Capabilities.*` instead?

cmd/result.go
```diff
@@ -44,25 +42,24 @@ func (res Result) Print() {
 	}
 }
 
-func createFields(res Result, err int) (fields log.Fields) {
+func createFields(res Result, occ Occurrence) (fields log.Fields) {
 	fields = log.Fields{}
 	v := reflect.ValueOf(res)
-	for _, member := range shouldLog(err) {
+	for _, member := range shouldLog(occ.id) {
```
klautcomputing: 👍

cmd/result.go
```diff
@@ -44,25 +42,24 @@ func (res Result) Print() {
...
 		value := v.FieldByName(member)
 		if value.IsValid() && value.Interface() != nil && value.Interface() != "" {
 			fields[member] = value.Interface()
 		}
 	}
+	for k, v := range occ.metadata {
```
klautcomputing: So `metadata` will always be logged, right? Should we add a comment/change the name then? Just so that it's purpose is obvious to other people that come and work on this repo.

cmd/result.go
fedorlis: `CapsAdded` and `CapsNotDropped` are no longer used since we have a single occurrence per erroneous cap (and we can use individual capNames). I might have gotten carried away deleting `CapsDropped` but we don't actually use it anywhere at the moment so I'm not sure what its there for

cmd/result.go
fedorlis: Metadata will only be logged if it is specified in the occurrence (otherwise `occ.metadata` will be nil and the for loop wont execute). 
Open to changing the name, I just picked metadata because its a store for all the miscellaneous fields that not all occurrences need to have. Maybe `fields`?

cmd/allowPrivilegeEscalation.go
```diff
@@ -4,17 +4,39 @@ import "github.com/spf13/cobra"
...
+		occ := Occurrence{
+			id:       ErrorAllowPrivilegeEscalationTrueAllowed,
+			kind:     Warn,
+			message:  "Allowed setting AllowPrivilegeEscalation to true",
+			metadata: map[string]string{"Reason": result.Labels["kubeaudit.allow.privilegeEscalation"]},
+		}
 		result.Occurrences = append(result.Occurrences, occ)
 		return
 	}
 	if container.SecurityContext.AllowPrivilegeEscalation == nil {
```
jinankjain: Since its a bool does it need explicit false value?

cmd/capabilities.go
```diff
@@ -46,40 +47,92 @@ func capsNotDropped(dropped []Capability) (notDropped []Capability, err error) {
...
+					message:  "Capability added",
+					metadata: map[string]string{"CapName": string(capAdded)},
+				}
+				result.Occurrences = append(result.Occurrences, occ)
+			} else {
+				occ := Occurrence{
+					id:      ErrorCapabilityAllowed,
+					kind:    Warn,
+					message: "Capability allowed",
+					metadata: map[string]string{
```
jinankjain: metadata could be a separate type Metadata which could have some json serialization associated with it.

cmd/result.go
klautcomputing: `fields` is fine, take a look at @jinankjain suggestion for making it its own `type` with json serialization.

jinankjain:
Hi @klautcomputing regarding inconsistencies I was discussing this with @fedorlis yesterday and it would be a cool feature for your autofix magic if it can detect that and fix those stuff automatically

fedorlis:
Discussed this just now, decided to throw an `ErrorMisconfiguredKubeauditAllow` warning if an allow label was used without reason.

klautcomputing:
Which doesn't mean I/we won't fit it automatically in autofix

jinankjain:
:100:

PR #75, opened 2017-12-20
Add autofix command

klautcomputing:


cmd/autofix.go
```diff
@@ -0,0 +1,159 @@
...
+
+func fixServiceAccountToken(resource k8sRuntime.Object) k8sRuntime.Object {
+	return setASAT(resource, false)
+}
+
+func fixDeprecatedServiceAccount(resource k8sRuntime.Object) k8sRuntime.Object {
+	return disableDSA(resource)
+}
+
+func fixStuff(resource k8sRuntime.Object, result Result) (fixed k8sRuntime.Object) {
```
jinankjain: A better name for this function like `fixPotentialSecurityIssues` or something similar

cmd/occurrence.go
```diff
@@ -1,7 +1,7 @@
 package cmd
 
 type Occurrence struct {
-	kind    int    // or int? just needs to represent  {debug, log, warn, error}
+	kind    int    // represent  {debug, log, warn, error}
```
jinankjain: may be using pesudo enum would be a good idea?

cmd/occurrence.go
```diff
@@ -1,7 +1,7 @@
...
 	id      int    // KubeAuditInfo, ErrorImageTagMissing ...
```
jinankjain: Same as above may be using pesudo enum

cmd/util.go
```diff
@@ -12,10 +12,21 @@ import (
...
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+func newTrue() *bool {
+	b := true
+	return &b
+}
+
+func newFalse() *bool {
```
jinankjain: using new(bool) would yield the same thing as this function

cmd/autofix.go
klautcomputing: @jinankjain It has the WIP prefix and label for a reason, there is a lot that needs to happen before this can be merged :) But I like the name you suggested.

cmd/util.go
klautcomputing: oh nice, I will totally do that!

cmd/autofix.go
jinankjain: @klautcomputing Yeah I was just going through PR. It seems super interesting feature. 💯

cmd/occurrence.go
klautcomputing: The `id` already an `enum` and defined in `errors.go`

cmd/occurrence.go
klautcomputing: This only holds an int which comes from the `loglevel.go` enum.

cmd/autofix.go
```diff
@@ -0,0 +1,159 @@
...
+func runAllAudits(resource k8sRuntime.Object) (results []Result) {
+	for _, function := range getAuditFunctions() {
+		for _, result := range getResults([]k8sRuntime.Object{resource}, function) {
+			results = append(results, result)
+		}
+	}
+	return results
+}
+
+func fixSecurityContextNIL(resource k8sRuntime.Object) k8sRuntime.Object {
```
fedorlis: I feel like this function should make an empty context and run each fix function on it, rather than using a set SecurityContext

cmd/capabilities_fixes.go
```diff
@@ -0,0 +1,39 @@
+package cmd
+
+import k8sRuntime "k8s.io/apimachinery/pkg/runtime"
+
+func fixCapabilitiesNIL(resource k8sRuntime.Object) k8sRuntime.Object {
```
fedorlis: Ditto here. Should we consider getting rid of `ErrorCapabilitiesNIL` all together and instead just throw a bunch of  `ErrorCapabilityNotDropped`s?

klautcomputing:
@jinankjain would you please give this another look so that we can get this merged once I have addressed yours and @fedorlis' comments. That's be awesome ❤️

cmd/test_util.go
```diff
@@ -73,3 +89,27 @@ func runAuditTestInNamespace(t *testing.T, namespace string, file string, functi
 	runAuditTest(t, file, function, errCodes)
 	rootConfig.namespace = apiv1.NamespaceAll
 }
+
+func NewPod() *Pod {
+	return &Pod{
```
jinankjain: It would good to use something `PodTemplate` to create a template for a pod

klautcomputing:
A readme update will follow once we implemented label support for `autofix`

PR #76, opened 2017-12-21
Refactor util.go, deletes all kubeaudit types

klautcomputing:
This is an attempt to switch away from all our internal types to `runtime.Objects`. Let me know what you think.

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -32,280 +32,105 @@ func isInNamespace(meta metav1.ObjectMeta, namespace string) (valid bool) {
 	return namespace == apiv1.NamespaceAll || namespace == meta.Namespace
 }
 
-func convertDeploymentToDeploymentList(deployment Deployment) (deploymentList *DeploymentList) {
-	if isInRootConfigNamespace(deployment.ObjectMeta) {
-		deploymentList = &DeploymentList{Items: []Deployment{deployment}}
-	} else {
-		deploymentList = &DeploymentList{Items: []Deployment{}}
+func getContainers(resource k8sRuntime.Object) (container []Container) {
+	switch kubeType := resource.(type) {
+	case *DaemonSet:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *Deployment:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *Pod:
+		container = kubeType.Spec.Containers
+	case *ReplicationController:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *StatefulSet:
+		container = kubeType.Spec.Template.Spec.Containers
+	}
+	return container
+}
+
+func newResultFromResource(resource k8sRuntime.Object) (result Result) {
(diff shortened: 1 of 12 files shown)


cmd/util.go
```diff
@@ -32,280 +32,96 @@ func isInNamespace(meta metav1.ObjectMeta, namespace string) (valid bool) {
...
+func getContainers(resource k8sRuntime.Object) (container []Container) {
+	switch kubeType := resource.(type) {
+	case *StatefulSet:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *DaemonSet:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *ReplicationController:
+		container = kubeType.Spec.Template.Spec.Containers
+	case *Deployment:
+		container = kubeType.Spec.Template.Spec.Containers
```
jinankjain: Since only pod is different than rest we can collapse all the case in the `default` behavior same thing can be done with other `switch` statements

jinankjain:
Finally our test coverage is coming close to 80%

cmd/util.go
klautcomputing: Unfortunately, that doesn't work because in a `default` you end up with an `runtime.Object` which doesn't have `Spec` only after type switching will the underlying methods and types be accessible.

```kubeType.Spec undefined (type "github.com/Shopify/kubeaudit/vendor/k8s.io/apimachinery/pkg/runtime".Object has no field or method Spec)```

I wish I had a more satisfying answer here for you. If you know of a way to make it work I will gladly do it that way.

File: cmd/util.go [hunk 2]
```diff
@@ -32,280 +32,106 @@ func isInNamespace(meta metav1.ObjectMeta, namespace string) (valid bool) {
...
+	}
+	return container
+}
+
+func newResultFromResource(resource k8sRuntime.Object) (result Result) {
+	switch kubeType := resource.(type) {
+	case *StatefulSet:
+		result.Name = kubeType.Name
+		result.Namespace = kubeType.Namespace
+		result.KubeType = kubeType.String()
```
fedorlis: This results in the whole object being printed, which is rather messy.
The best alternative I could come up with is `result.KubeType = strings.Split(reflect.TypeOf(kubeType).String(), ".")[1]`

cmd/util.go
```diff
@@ -32,280 +32,106 @@ func isInNamespace(meta metav1.ObjectMeta, namespace string) (valid bool) {
...
-
-type kubeAuditPods struct {
-	list *PodList
-}
-
-type kubeAuditReplicationControllers struct {
-	list *ReplicationControllerList
-}
-
-type Items interface {
```
fedorlis: ❤️

cmd/util.go
klautcomputing: Good catch, I will just go with the old solution and put the strings back in, since we need a case for every type there is no need for the `reflect` here.

PR #79, opened 2018-01-05
Add stricter check for audit tests

fedorlis:
This PR enables our tests to ensure that only errors that were specified were thrown.

### DIFF
--- a/cmd/test_util.go
+++ b/cmd/test_util.go
@@ -12,7 +12,7 @@ import (
 
 var path = "../fixtures/"
 
-func runTest(t *testing.T, file string, function interface{}, errCode int, argStr ...string) (results []Result) {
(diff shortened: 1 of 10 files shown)


cmd/test_util.go
```diff
@@ -54,21 +54,22 @@ func runTest(t *testing.T, file string, function interface{}, errCode int, argSt
...
 	for _, result := range results {
 		for _, occurrence := range result.Occurrences {
-			errors = append(errors, occurrence.id)
+			errors[occurrence.id] = true
 		}
 	}
 
-	if errCode != 0 {
-		assert.Contains(errors, errCode)
+	for _, errCode := range errCodes {
```
klautcomputing: Nitpick: I would swap the two asserts because if the length is off you know it's not gonna work :)

Issue #84, opened 2018-01-12
Enhance autofix's yaml handling


klautcomputing:
When running autofix it reads the yaml file correctly, the resource gets fixed and the write works as well. Yet, there are 3 things that could be improved upon:

## Empty resources get added to the yaml file: 

```yaml
status:
  loadBalancer: {}
```

## Yaml comments disappear

This is a know issue in Yaml and has been an open issue in go-yaml for more than 2 years, see:
- https://github.com/go-yaml/yaml/issues/132

There is an initial version of a patch out there but it was never finished:
- https://github.com/DirectXMan12/go-yaml/commit/4c95efea86311a44556aad1cc58f449bb5b7e65f

Seems like there is one parser in python that can preserve comments
- https://pypi.python.org/pypi/ruamel.yaml

## Order is not preserved

Support for `yaml.MapSlice` was added in `go-yaml.v2` see:
- https://blog.labix.org/2014/09/22/announcing-yaml-v2-for-go

The whole discussion about the feature can be found here 
 - https://github.com/go-yaml/yaml/issues/30

PR #89, opened 2018-01-31
Fixes #87 autofix

### LINKED ISSUE #87: autofix only writes back the last resource
Right now autofix uses this part of the code to write back the resources to file:
https://github.com/Shopify/kubeaudit/blob/master/cmd/k8sruntime_util.go#L96-L109
Which means only the last resources is actually in the file afterwards.

natalysheinin:
Fix for this issue: https://github.com/Shopify/kubeaudit/issues/87

### DIFF
--- a/cmd/k8sruntime_util.go
+++ b/cmd/k8sruntime_util.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"io/ioutil"
+	"os"
 
 	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


cmd/k8sruntime_util.go
```diff
@@ -93,17 +94,30 @@ func getContainers(resource k8sRuntime.Object) (container []Container) {
...
-	if err != nil {
-		return err
+	if !toAppend {
+		err = ioutil.WriteFile(filename, yaml, 0644)
+		if err != nil {
+			return err
+		}
+	} else {
+		f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+		defer f.Close()
```
klautcomputing: The `defer f.Close()` call needs to be moved to after the `err != nil` block because otherwise you might try to close on a non existent `f`

klautcomputing:
Another general tip if you use one of [these](https://help.github.com/articles/closing-issues-using-keywords/) keywords before the #number it will automatically close the issue when this gets merged.

klautcomputing:
#84

PR #90, opened 2018-02-02
Fixes #88

### LINKED ISSUE #88: 2x your capapility drops!
Currently, autofix does not detect that caps have already been dropeed, so it drops them again.
I haven't had a look at why, but this is the result:

```yaml
          capabilities:
            drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - SETGID
            - SETFCAP
            - SETPCAP
            - SETUID
            - SYS_CHROOT
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETPCAP
            - SETUID
            - SYS_CHROOT
```

klautcomputing:
Kubeaudit now correctly audits all containers.

closes #88

### DIFF
--- a/cmd/capabilities_fixes.go
+++ b/cmd/capabilities_fixes.go
@@ -22,7 +22,9 @@ func fixCapabilitiesNIL(resource k8sRuntime.Object) k8sRuntime.Object {
 func fixCapabilityNotDropped(resource k8sRuntime.Object, occurrence Occurrence) k8sRuntime.Object {
 	var containers []Container
 	for _, container := range getContainers(resource) {
-		container.SecurityContext.Capabilities.Drop = append(container.SecurityContext.Capabilities.Drop, Capability(occurrence.metadata["CapName"]))
+		if occurrence.container == container.Name {
+			container.SecurityContext.Capabilities.Drop = append(container.SecurityContext.Capabilities.Drop, Capability(occurrence.metadata["CapName"]))
+		}
 		containers = append(containers, container)
 	}
 	return setContainers(resource, containers)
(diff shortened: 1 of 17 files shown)


cmd/capabilities_fixes.go
```diff
@@ -14,31 +14,32 @@ func fixCapabilitiesNIL(resource k8sRuntime.Object) k8sRuntime.Object {
...
 }
 
 func fixCapabilityNotDropped(resource k8sRuntime.Object, occurrence Occurrence) k8sRuntime.Object {
-	var containers []Container
-	for _, container := range getContainers(resource) {
-		container.SecurityContext.Capabilities.Drop = append(container.SecurityContext.Capabilities.Drop, Capability(occurrence.metadata["CapName"]))
-		containers = append(containers, container)
+	containers := getContainers(resource)
+	for _, container := range containers {
+		if occurrence.container == container.Name {
```
jinankjain: Is container.name unique for every container inside a k8s resource?

fedorlis: According to https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/, names are unique within namespaces but not necessarily across them.

klautcomputing: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2089-L2093 So from what I take from the comment they should be unique.

klautcomputing:
As @jinankjain pointed out the code currently parses invalid `yaml` files incorrectly instead of erroring out.
I don't think that's really in our realm to address general validation of templates. I really like the Unix philosophy of: *Do one thing do it well*. In our case that would be address security issues. If the config is not correct that should be caught by a validator.
But I will patch a quick check in there that names need to be unique :)

cmd/util_test.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDoubleName(t *testing.T) {
```
fedorlis: ❤️

klautcomputing:
#84

PR #111, opened 2018-09-27
capabilities: Fix a false positive when all capabilities dropped

### LINKED ISSUE #110: False positive when all capabilities dropped
Running `kubeaudit caps` returns a lot of `ERRO[0003] Capability not dropped` messages for the pods with the following effective settings:
```yaml
    securityContext:
      capabilities:
        drop:
        - all
      readOnlyRootFilesystem: true
```
From inside the container everything is clearly OK:
```
grep ^Cap /proc/1/status
CapInh:    0000000000000000
CapPrm:    0000000000000000
CapEff:    0000000000000000
CapBnd:    0000000000000000
CapAmb:    0000000000000000
```

jinankjain:
Fixes #110 

@Tmlja1Jldmlu hopefully this fix the false positive that you were observing.

Thanks for filing the issue. 😄

### DIFF
--- a/cmd/capabilities.go
+++ b/cmd/capabilities.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"io/ioutil"
 	"os"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/capabilities.go
```diff
@@ -60,9 +60,14 @@ func recommendedCapabilitiesToBeDropped() (dropCapSet CapSet, err error) {
 func checkCapabilities(container Container, result *Result) {
 	added := CapSet{}
 	dropped := CapSet{}
+	allCapsDrop := false
 	if container.SecurityContext != nil && container.SecurityContext.Capabilities != nil {
 		added = NewCapSetFromArray(container.SecurityContext.Capabilities.Add)
-		dropped = NewCapSetFromArray(container.SecurityContext.Capabilities.Drop)
+		if len(container.SecurityContext.Capabilities.Drop) != 0 && string(container.SecurityContext.Capabilities.Drop[0]) == "all" {
```
Tmlja1Jldmlu: Thanks for the quick fix!

It will be better to `strings.ToLower(string(container.SecurityContext.Capabilities.Drop[0]))`.

Unfortunately I cannot find proper refs anywhere in k8s docs or code but my tests show that this `ALL` keyword is case insensitive...

Issue #112, opened 2018-09-27
Version Command Panics


shane-lawrence:
`kubeaudit version` returns the wrong version number and then panics.

```
INFO[0000] Kubeaudit                                     Version=0.1.0
Running inside cluster, using the cluster config
ERRO[0000] unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1059878]
```

jinankjain:
@shane-lawrence one solution could be that by default set the value of local to be true and use special flag to indicate its inside a cluster. 

On a side note I think we need to change that version string too.

shane-lawrence:
The version string needs to be updated and I noticed a few times that the kubectl client version was blank so I'll look into that as well.

Here's my proposed fix for the panic: https://github.com/Shopify/kubeaudit/pull/113

shane-lawrence:
Resolved by https://github.com/Shopify/kubeaudit/pull/113.

PR #113, opened 2018-09-28
Automatically detect local/cluster mode

shane-lawrence:
This changes the default behaviour of kubeaudit so that it will automatically use the default local config from `$HOME/.kube/config` any time it's not running within a k8s container. You can still force it to use local config mode if it is running in a cluster by specifying `-c/--kubeconfig /config/path`.

The `-l` flag is no longer necessary, but you can still use it and it will behave as before so it doesn't break backwards compatibility. It will display a warning if you use it.

I moved the logging initialization to root so that formatting and level will be respected for all commands even if you don't call kubeClient(). 

Edit: fixing the version number and adding tests will be done in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,17 +46,24 @@ Now you can just call `kubeaudit` with one of commands from [here](#audits)
 ## General instructions
 
 `kubeaudit` has three different modes for its audits:
-1. `kubeaudit cmd` will attempt to create an in-cluster client and audit.
-1. `kubeaudit -l/--local cmd` will use your kubeconfig (`~/.kube/config` or if
-   you need different path use `-c /config/path`
-1. `kubeaudit -f/--manifest /path/to/manifest.yml` will audit the manifest
+1. Cluster mode
+  If kubeaudit detects that it's running in a container, `kubeaudit cmd` will
+  attempt to audit the cluster it's running in.
+1. Local config mode
+  If kubeaudit is not running in a container, `kubeaudit cmd` will audit the
+  resources specified by your local kubeconfig (`$HOME/.kube/config`) file.
+  You can force kubeaudit to use a specific local config file with the switch
+  `-c/--kubeconfig /config/path`
+1. Manifest mode
+  If you wish to audit a manifest file, use the command
+  `kubeaudit -f/--manifest /path/to/manifest.yml`
 
-`kubeaudit` supports to different output types:
+`kubeaudit` supports two different output types:
 1. just running `kubeaudit` will log human readable output
 1. if run with `-j/--json` it will log output json formatted so that its output
    can be used by other programs easily
 
-`kubeaudit` has 4 different log levels `INFO, WARN, ERROR` controlled by
+`kubeaudit` has four different log levels `INFO, WARN, ERROR` controlled by
 `-v/--verbose LEVEL` and for those who counted and want to work on `kubeaudit`
 `DEBUG`
 1. by default the debug level is set to `ERROR` and will log `INFO`, `WARN` and
@@ -105,7 +112,7 @@ The manifest might end up a little too secure for the work it is supposed to do.
 Runs all the above checks.
 
 ```sh
-kubeaudit -l all
+kubeaudit all
 ERRO[0000] RunAsNonRoot is not set, which results in root user being allowed!
 ERRO[0000] Default serviceAccount with token mounted. Please set automountServiceAccountToken to false
 WARN[0000] Privileged defaults to false, which results in non privileged, which is okay.
@@ -120,7 +127,7 @@ The security context holds a couple of different security related
 configurations. For convenience, `kubeaudit` will always log the following
 information when it creates a log:
 ```sh
-kubeaudit -l command
+kubeaudit command
 LOG[0000] KubeType=deployment Name=THEdeployment Namespace=deploymentNS
 ```
 And for brevity, the information will not be shown in the commands below.
@@ -134,7 +141,7 @@ Currently, `kubeaudit` is able to check for the following fields in the security
 `kubeaudit` will detect whether `readOnlyRootFilesystem` is either not set `nil` or explicitly set to `false`
 
 ```sh
-kubeaudit -l rootfs
+kubeaudit rootfs
 ERRO[0000] ReadOnlyRootFilesystem not set which results in a writable rootFS, please set to true
 ERRO[0000] ReadOnlyRootFilesystem set to false, please set to true
 ```
@@ -146,7 +153,7 @@ ERRO[0000] ReadOnlyRootFilesystem set to false, please set to true
 `kubeaudit` will detect whether the container is to be run as root:
 
 ```sh
-kubeaudit -l nonroot
+kubeaudit nonroot
 ERRO[0000] RunAsNonRoot is set to false (root user allowed), please set to true!
 ERRO[0000] RunAsNonRoot is not set, which results in root user being allowed!
 ```
@@ -158,7 +165,7 @@ ERRO[0000] RunAsNonRoot is not set, which results in root user being allowed!
 `kubeaudit` will detect whether `allowPrivilegeEscalation` is either set to `nil` or explicitly set to `false`
 
 ```sh
-kubeaudit -l allowpe
+kubeaudit allowpe
 ERRO[0000] AllowPrivilegeEscalation set to true, please set to false
 ERRO[0000] AllowPrivilegeEscalation not set which allows privilege escalation, please set to false
 ```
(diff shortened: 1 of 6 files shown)


jonpulsifer:
hi my name is jon and i approve WIP PRs

cmd/util.go
```diff
@@ -36,6 +37,11 @@ func debugPrint() {
 	}
 }
 
+func isInContainer() bool {
+	host, port := os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")
```
jinankjain: It would be a good idea to make those strings ("KUBERNETES_SERVICE_HOST", "KUBERNETES_SERVICE_PORT") as `const`

It would be good to use `os.LookupEnv` instead of using `os.Getenv` and checking the length

cmd/version.go
```diff
@@ -28,12 +29,7 @@ var versionCmd = &cobra.Command{
...
 
-		kubeconfig := rootConfig.kubeConfig
-		if rootConfig.localMode {
-			kubeconfig = os.Getenv("HOME") + "/.kube/config"
-		}
-
-		kube, err := kubeClient(kubeconfig)
+		kube, err := kubeClient()
 		if err != nil {
 			log.Error(err)
```
jinankjain: may be log.Panic() since we are anyways exiting after that

cmd/kubernetes.go
```diff
@@ -14,26 +13,22 @@ import (
...
 )
 
-func kubeClientConfig(kubeconfig string) (*rest.Config, error) {
-	if kubeconfig == "" && !rootConfig.localMode {
-		fmt.Println("Running inside cluster, using the cluster config")
-		return rest.InClusterConfig()
+func kubeClientConfig() (*rest.Config, error) {
+	if rootConfig.kubeConfig == "" && !isInContainer() {
+		log.Info("Not running inside cluster, using local config")
+		rootConfig.kubeConfig = os.Getenv("HOME") + "/.kube/config"
```
jinankjain: `filepath.Join()` would be a better idea than simple string concatenation

shane-lawrence:
I incorporated most of your suggestions and decided to let the client decide if we’re running in a cluster or not. Lmk what you think.

README.md
```diff
@@ -191,14 +198,14 @@ therefore should be dropped. `kubeaudit` will also complain about added capabili
 If the capabilities field doesn't exist within the security context:
 
 ```sh
-kubeaudiit -l caps
+kubeaudiit caps
```
jonpulsifer: kubeaudit

shane-lawrence:
Wow I can’t believe I missed that.

cmd/root.go
```diff
@@ -41,12 +42,31 @@ func Execute() {
...
+		log.AddHook(NewDebugHook())
+	}
+	if rootConfig.json {
+		log.SetFormatter(&log.JSONFormatter{})
+	}
+
+	if rootConfig.localMode == true {
+		log.Warn("-l/-local is deprecated! kubeaudit will default to local checks if it's not running in a cluster. ")
+		log.Warn("To use a local kubeconfig file from inside a cluster specify '-c $HOME/.kube/config'.")
+		rootConfig.kubeConfig = os.Getenv("HOME") + "/.kube/config"
```
jinankjain: filepath.Join()

jinankjain:
One more thing which was missing was the version string update from 0.1 to 0.3

cmd/root.go
```diff
@@ -41,12 +42,31 @@ func Execute() {
 }
 
 func init() {
-	RootCmd.PersistentFlags().StringVarP(&rootConfig.kubeConfig, "kubeconfig", "c", "", "config file (default is $HOME/.kube/config")
+	cobra.OnInitialize(processFlags)
+	RootCmd.PersistentFlags().BoolVarP(&rootConfig.localMode, "local", "l", false, "[DEPRECATED] Local mode, uses $HOME/.kube/config as configuration")
+	RootCmd.Flags().MarkHidden("local")
+
```
klautcomputing: s/\\n//

cmd/root.go
klautcomputing: https://golang.org/pkg/os/#LookupEnv is better than `GetEnv`, well I admit "Home" should always be there, but generally it's the function, because it checks for empty and returns a bool that tells you whether the variable exists or not.

jinankjain: Yeah sounds like a better safety check 👍

cmd/version.go
```diff
@@ -28,15 +29,9 @@ var versionCmd = &cobra.Command{
...
-		if rootConfig.localMode {
-			kubeconfig = os.Getenv("HOME") + "/.kube/config"
-		}
-
-		kube, err := kubeClient(kubeconfig)
+		kube, err := kubeClient()
 		if err != nil {
-			log.Error(err)
-			os.Exit(1)
+			log.Panic(err)
```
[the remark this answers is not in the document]
shane-lawrence: I dropped that commit and added this one instead https://github.com/Shopify/kubeaudit/pull/113/commits/bcbbdc8523f7c9ca07bd231bc39665e45bf7228e.

cmd/root.go
shane-lawrence: Resolved in https://github.com/Shopify/kubeaudit/pull/113/commits/9c8b9b6a7958a9338e757e4e25fe8f295b217957.

cmd/kubernetes.go
```diff
@@ -14,26 +14,24 @@ import (
...
+		log.Info("Running inside cluster, using the cluster config")
+		return config, nil
 	}
-	if rootConfig.json {
-		log.SetFormatter(&log.JSONFormatter{})
-	}
-	config, err := kubeClientConfig(kubeconfig)
+
+	log.Info("Not running inside cluster, using local config")
+	rootConfig.kubeConfig = filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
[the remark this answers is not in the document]
shane-lawrence: https://github.com/Shopify/kubeaudit/pull/113/commits/466c69208a7cc5ccb0e915c8a648961c2f29a924 adds handling for missing kubeconfig file.

cmd/root.go
```diff
@@ -41,12 +40,34 @@ func Execute() {
...
+func processFlags() {
+	if rootConfig.verbose == "DEBUG" {
+		log.SetLevel(log.DebugLevel)
+		log.AddHook(NewDebugHook())
+	}
+	if rootConfig.json {
+		log.SetFormatter(&log.JSONFormatter{})
+	}
+
+	if rootConfig.localMode == true {
```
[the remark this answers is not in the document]
shane-lawrence: That's how kubeaudit handled `-l` prior to my changes. Here is a sample from the current master branch:
```
./kubeaudit -l version
INFO[0000] Kubeaudit                                     Version=0.1.0
INFO[0000] Kubernetes server version                     Major=1 Minor=10+ Platform=linux/amd64
INFO[0000] Kubernetes client version                     Major= Minor= Platform=darwin/amd64

./kubeaudit -l -c /notarealfile version
INFO[0000] Kubeaudit                                     Version=0.1.0
INFO[0000] Kubernetes server version                     Major=1 Minor=10+ Platform=linux/amd64
INFO[0000] Kubernetes client version                     Major= Minor= Platform=darwin/amd64
```

Since I only kept -l in for backwards compatibility, I thought it best to maintain the existing behaviour. However, I agree that it's not the behaviour one would expect and that it would be improved by having `-c` override `-l`.

It's a trivial fix so just let me know if you still think I should change it.

shane-lawrence:
https://github.com/Shopify/kubeaudit/pull/113/commits/5a4f94febca178e36df88f3d31bf6b29473b8b0d logs errors describing why the client failed. For example, if the kubeconfig file can't be found the output will look like this:
```
./kubeaudit -c /notarealfile version
INFO[0000] Kubeaudit                                     Version=0.1.0
ERRO[0000] Unable to load kubeconfig. Could not open file /notarealfile.
WARN[0000] Could not get kubernetes server version.
```

cmd/root.go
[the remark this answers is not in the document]
klautcomputing: I think if we are already touching the code it would be cool to fix this bug :) But yeah I can see where you are coming from and honestly I didn't know it behaved like that.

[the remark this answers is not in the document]
shane-lawrence: Fixed in https://github.com/Shopify/kubeaudit/pull/113/commits/4f678c4e97f5e651298dd5f0cbd6192fe41c9bf0. This is the new behaviour:
```
./kubeaudit -l -c /notarealfile version
WARN[0000] -l/-local is deprecated! kubeaudit will default to local mode if it's not running in a cluster.
INFO[0000] Kubeaudit                                     Version=0.1.0
ERRO[0000] Unable to load kubeconfig. Could not open file /notarealfile.
WARN[0000] Could not get kubernetes server version.
```

shane-lawrence:
Fixes #112 and #116.

PR #132, opened 2018-11-02
Update dependencies

shane-lawrence:
This updates dependencies and moves Docker builds to Go 1.11. It adds tests for both Go 1.10 and 1.11. 

The kubernetes client is no longer compatible with Go 1.9 due to https://github.com/kubernetes/client-go/blob/02384dbe123ff097a279965297f327a72ebefb72/transport/round_trippers.go#L437. I've added a check to the Makefile. This is what you'll see if you try to build with Go 1.9 or earlier:
https://travis-ci.org/Shopify/kubeaudit/builds/450992200#L468

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: go
 
 go:
-  - 1.9.1
+  - "1.10.x"
+  - "1.11.x"
 
 script:
   - make setup
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,20 +2,20 @@
 
 
 [[projects]]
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: go
 
 go:
-  - 1.9.1
+  - "1.9.x"
```
klautcomputing:
```suggestion
```

shane-lawrence:
I wasn't able to replicate the issue, but it seems to have been caused by some weirdness with the vendor directory. `rm -rf vendor/` will eliminate the problem. Please let me know if it recurs.

shane-lawrence:
I edited the PR comment to summarize the changes.

PR #137, opened 2018-12-04
Rename kubernetes type aliases to include versions

genevieveluyt:
Adds the version to all of the kubernetes type aliases. If there is a conflict (ie. a type is present in the same version of two different groups), then the type name includes both the group and the version.

### DIFF
--- a/cmd/types.go
+++ b/cmd/types.go
@@ -2,102 +2,102 @@ package cmd
 
 import (
 	appsv1 "k8s.io/api/apps/v1"
-	v1beta1 "k8s.io/api/apps/v1beta1"
-	v1beta2 "k8s.io/api/apps/v1beta2"
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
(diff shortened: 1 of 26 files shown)


cmd/types.go
```diff
@@ -2,112 +2,112 @@ package cmd
...
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	apiv1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
-	networking "k8s.io/api/networking/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// CronJob is a type alias for the v1beta1 version of the k8s API.
```
genevieveluyt: The types are in alphabetical order now

PR #140, opened 2018-12-20
Add comment support for autofix

### LINKED ISSUE #84: Enhance autofix's yaml handling

### LINKED ISSUE #92: Autofixing resources deletes yaml document separators
Problem: `---` disappear when using `--autofix`
Solution: Add them back in

ref #84

### LINKED ISSUE #146: Autofix skips and drops resources
When kubeaudit is run with `autofix -f file.yaml` and the to be autofixed file contains resources that kubeaudit doesn't know  about like e.g. ingress and service the following happens:
```
WARN[0000] Skipping unsupported resource type extensions/v1beta1, Kind=Ingress
WARN[0000] Skipping unsupported resource type /v1, Kind=Service
```
Kubeaudit skips and drops them. What it should do is skip and keep.

### LINKED ISSUE #152: Extra newline generated by autofix on manifest starting with comment after yaml separator
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

# BUG REPORT

##### SUMMARY
Extra newline is generated by Autofix on manifest starting with comment after yaml separator.
##### ENVIRONMENT
* Kubeaudit version: 0.4.1 (branch autofix)
* Kubeaudit install method: -

##### STEPS TO REPRODUCE

Create a manifest file with the following structure
```
---
#This is a comment 3
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
  creationTimestamp: null
spec:
  rules: #This is a comment 1
  - http:
      paths:
      - backend:
          serviceName: test
          servicePort: 80
        path: /testpath
status:
  loadBalancer: {}
#This is a comment 5
```

run
```
kubeaudit autofix -f /path/to/manifest.yml
```

##### EXPECTED RESULTS

There should not be extra newline after yaml separator

##### ACTUAL RESULTS

changes the file to  
```
---
  
#This is a comment 3
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
  creationTimestamp: null
spec:
  rules: #This is a comment 1
  - http:
      paths:
      - backend:
          serviceName: test
          servicePort: 80
        path: /testpath
status:
  loadBalancer: {}
#This is a comment 5
```

genevieveluyt:
This PR makes it so that the autofix command will preserve the order and comments of the input YAML file.

## How it works

img_20181220_131045

The input file, `workload.yaml` is autofixed as it previously was. We then take this YAML and merge it back into the original `workload.yaml`.

To merge the two YAML files, we first convert them into MapSlices, which is a map represented as an array so as to preserve the order of the map items. This is done using the `Shopify/yaml` library. We then recursively merge the MapSlices. At each recursive call, three things can happen:
* Items can be added
* Items can be removed
* Items can be updated

First we look for any items that exist in the original YAML but not the fixed YAML and remove them. That is, for each item in the original for which there is no "matching" item in the fixed, we remove that item ("match" is defined below). Then, for each item in the fixed YAML we check if there is a "match" for it in the original. If there is, we update the item, otherwise we add the item.

### Matching
There are three different matching cases:
* MapSlice
* SequenceItem array
* primitive value

A MapSlice item has a Key, a Value, and a Comment. To determine whether MapSlice items are a match we simply compare the keys. If the keys are the same, they are a match.

A SequenceItem has only a Value and a Comment so we determine whether they are a match based on their Value. If the Value is a primitive type, we compare them directly, but if not then determining whether SequenceItems are a match is a bit more complicated. Instead of going with the general case of determining matches for any arbitrary valid YAML, we can simplify the problem by using what we know about kubernetes config files. For example:
```
containers:
- name: container1
  securityContext:
    capabilities:
      drop:
      - AUDIT_WRITE
- name: container2
```
If the list key is "containers", then each list item must contain a "name" key (the "identifying key") with a unique value within that list, so we can determine a SequenceItem match based on whether they have the same key-value pair for this "identifying key". These "list key" -> "identifying key" pairs are compiled manually based on the kubernetes api.

Primitive values are compared directly.

Trying to match values of different types will always be considered a non-match.
This closes #146 , closes #92 , closes #84, closes #152


klautcomputing:
@jinankjain would you do me the favor and take a look at it as well?

fedorlis:
It's amazing to see this finally merge, thank you for doing this! I've 🎩 'd and it seems to work like a treat, but I don't think I have enough context on the codebase anymore to properly review this.

cmd/autofix.go
```diff
@@ -86,15 +89,249 @@ func fix(resources []k8sRuntime.Object) (fixedResources []k8sRuntime.Object) {
...
+			continue
+		}
+		if findKeyInMapSlice(item.Key, fixedSlice) == -1 {
+			origSlice = append(origSlice[:i], origSlice[i+1:]...)
+			i--
+		}
+	}
+
+	// Update or add items from the fixed yaml which are not in the original
+	for i := 0; i < len(fixedSlice); i++ {
```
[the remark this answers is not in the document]
genevieveluyt: I did a bunch of refactoring and forgot to clean this up. Will fix!

fixtures/autofix_v1.yml
```diff
@@ -1,22 +1,34 @@
...
+
+---
+
+apiVersion: apps/v1 # The latest api version for Deployment
+# This is a deployment
 kind: Deployment
 metadata:
-  creationTimestamp: null
-  name: cababilitiesAdded
+  creationTimestamp: null # This is autopopulated
```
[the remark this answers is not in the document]
genevieveluyt: It was already there, I just added a random comment. Is it less confusing if I change the comment text?

cmd/autofix_test.go
```diff
@@ -31,3 +33,28 @@ func TestFixV1Beta1(t *testing.T) {
 	assert.Nil(err)
 	assert.Nil(deep.Equal(correctlyFixedResources, fixedResources))
 }
+
+func TestPreserveComments(t *testing.T) {
+	origFile := "../fixtures/autofix_v1.yml"
```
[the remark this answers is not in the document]
genevieveluyt: Originally I didn't want to do a text comparison because I thought that would be too specific and assumed that was also the reason the existing tests compared resources instead of file contents, but considering a bug got through, maybe text comparisons would be better for all the tests? I will change this one to call autofix and compare the file contents (it does pass).

cmd/autofix.go
```diff
@@ -21,32 +21,51 @@ func getAuditFunctions() []interface{} {
...
+
 	resources, err := getKubeResourcesManifest(rootConfig.manifest)
 	if err != nil {
 		log.Error(err)
 	}
 
-	fixedResources := fix(resources)
-
-	tmpFile, err := ioutil.TempFile("", "kubeaudit_autofix")
+	err = os.Truncate(rootConfig.manifest, 0)
```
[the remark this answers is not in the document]
nschhina: Great catch! Should definitely do this!

cmd/autofix.go
```diff
@@ -21,32 +21,51 @@ func getAuditFunctions() []interface{} {
 // preserves the order of the keys) using the Shopify/yaml fork of go-yaml/yaml (the fork adds comment support) and
 // then merge the fixed MapSlice back into the original MapSlice so that we get the comments and original order back.
 func autofix(*cobra.Command, []string) {
+
+	var toAppend bool
```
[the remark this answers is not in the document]
nschhina: I didn't explicitly set it to False since nil value for Bool is false and I don't think we were setting toAppend to false in the earlier version, but I guess it wouldn't hurt to explicitly set it to false :)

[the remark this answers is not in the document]
genevieveluyt: This is purely opinion based and I have definitely heard arguments for both sides. Personally, I think explicitly setting values even for the default improves readability but feel free to leave it as is if you like

cmd/autofix.go
```diff
@@ -13,87 +16,65 @@ func getAuditFunctions() []interface{} {
...
-		}
-	}
+	fixedResources := fix(resources)
 
-	for _, err := range needCapabilitiesDefined {
-		if _, ok := errors[err]; ok {
-			resource = fixCapabilitiesNil(resource)
-			break
-		}
+	tmpFixedFile, err := ioutil.TempFile("", "kubeaudit_autofix_fixed")
```
genevieveluyt: `defer os.Remove(tmpFixedFile.Name())` ? The other files too

cmd/autofix.util.go
```diff
@@ -0,0 +1,531 @@
...
+//     []yaml.SequenceItem{{Value: yaml.MapSlice{
+// 	       {Key: "k", Value: "v", Comment: "c"},
+// 	       {Key: "k2", Value: "v2", Comment: "c2"},
+//     }}}
+//
+//     []yaml.SequenceItem{{Value: yaml.MapSlice{
+//          {Key: "k2", Value: "v2"},
+//          {Key: "k", Value: "v"},
+//      }}}
+
```
klautcomputing:
```suggestion
```

cmd/types.go
```diff
@@ -81,6 +81,7 @@ type ReplicationControllerListV1 = apiv1.ReplicationControllerList
 // ReplicationControllerV1 is a type alias for the v1 version of the k8s API.
 type ReplicationControllerV1 = apiv1.ReplicationController
 
+// Resource is a type alias for a k8s Object.
```
klautcomputing: well to be precise for `runtime.Object`. Not sure what a k8s Object is :)

cmd/util.go
```diff
@@ -206,14 +207,24 @@ func getKubeResources(clientset *kubernetes.Clientset) (resources []Resource) {
...
+func writeManifestFile(decoded []byte, filename string, toAppend bool) error {
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	if toAppend {
+		f.WriteString("---\n")
+	}
+	defer f.Close()
```
klautcomputing: I would move the defer up a little.

fixtures/autofix-unsupported_v1.yml
```diff
@@ -0,0 +1,72 @@
...
+metadata:
+  name: test-ingress
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: /
+  creationTimestamp: null
+spec:
+  rules:
+  - http:
+      paths:
+      - backend: #This is a comment 4
```
klautcomputing: When running autofix, this comment gets swallowed.

nschhina: https://github.com/Shopify/kubeaudit/blob/autofix/cmd/autofix.util.go#L415 http was missing in ports case, has been added. Works like a charm now

cmd/util.go
```diff
@@ -213,10 +213,12 @@ func writeManifestFile(decoded []byte, filename string, toAppend bool) error {
 		log.Error(err)
 		return err
 	}
+	defer f.Close()
 	if toAppend {
 		f.WriteString("---\n")
 	}
-	defer f.Close()
+	// Remove newline from the decoded slice if the slice starts with a newline
+	decoded = []byte(strings.TrimPrefix(string(decoded), "\n"))
```
genevieveluyt: Optionally, if you want to get around converting twice
```
if decoded[0] == '\n' {
     decoded = byte[1:]
}
```

nschhina: I thought TrimPrefix was a cleaner approach and very little performance difference. @klautcomputing can you weigh in?

klautcomputing: tbh: I don't care too much either way.

PR #144, opened 2019-01-04
Audit all containers

### LINKED ISSUE #141: Autofix broken for multiple containers
When autofixing

```
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  name: cababilitiesAdded
  namespace: fakeDeploymentSC
spec:
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        apps: fakeSecurityContext
    spec:
      containers:
      - name: fakeContainerSC1
        securityContext:
          capabilities:
            drop:
            - AUDIT_WRITE
      - name: fakeContainerSC2
```
The resulting YAML is
```
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  name: cababilitiesAdded
  namespace: fakeDeploymentSC
spec:
  selector: null
  strategy: {}
  template:
    metadata:
      annotations:
        container.apparmor.security.beta.kubernetes.io/fakeContainerSC1: runtime/default
        container.apparmor.security.beta.kubernetes.io/fakeContainerSC2: runtime/default
        seccomp.security.alpha.kubernetes.io/pod: runtime/default
      creationTimestamp: null
      labels:
        apps: fakeSecurityContext
    spec:
      automountServiceAccountToken: false
      containers:
      - name: fakeContainerSC1
        resources: {}
        securityContext:
          capabilities:
            drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETPCAP
            - SETUID
            - SYS_CHROOT
          privileged: false
          readOnlyRootFilesystem: true
          runAsNonRoot: true
      - name: fakeContainerSC2
        resources: {}
        securityContext:
          capabilities:
            drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETPCAP
            - SETUID
            - SYS_CHROOT
status: {}
```
which has
```
privileged: false
readOnlyRootFilesystem: true
runAsNonRoot: true
```
added for the first container but not the second. The example above is actually the test file `fixtures/autofix_v1.yml`. It tests against the expected output `fixtures/autofix-fixed_v1.yml`. The expected output has
```
privileged: false
readOnlyRootFilesystem: true
runAsNonRoot: true
```
for both containers yet the test still passes...

klautcomputing:
This fixes #141 

Turns out if you `break` out of your loop then you never get shit done 🙂 

Added a regression test. Turns out: deep equal from `deep` doesn't catch the difference, deep equal from `reflect` complains that they are not equal, so I serialize them, write them to file and compare the file contents.

### DIFF
--- a/cmd/autofix_test.go
+++ b/cmd/autofix_test.go
@@ -1,49 +1,10 @@
 package cmd
 
 import (
-	"bufio"
-	"fmt"
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
-
-func compareFiles(file1, file2 string) bool {
-	f1, err := os.Open(file1)
-	if err != nil {
-		return false
-	}
-
-	f2, err := os.Open(file2)
-	if err != nil {
-		return false
-	}
-
-	s1 := bufio.NewScanner(f1)
-	s2 := bufio.NewScanner(f2)
-
-	for s1.Scan() {
-		s2.Scan()
-		text1 := s1.Text()
-		text2 := s2.Text()
-		if text1 != text2 {
-			fmt.Printf("Files don't match here:\n%v\n%v\n\n", text1, text2)
-			return false
-		}
-	}
-	return true
-}
-
-func assertEqualWorkloads(assert *assert.Assertions, resource1, resource2 []Resource) {
-	tmpfile1, err := WriteToTmpFile(resource1[0])
-	assert.Nil(err)
-	defer os.Remove(tmpfile1)
-	tmpfile2, err := WriteToTmpFile(resource2[0])
-	assert.Nil(err)
-	defer os.Remove(tmpfile2)
-	assert.True(compareFiles(tmpfile1, tmpfile2))
-}
 
 func TestFixV1(t *testing.T) {
 	file := "../fixtures/autofix_v1.yml"
--- a/cmd/autofix_test.go
+++ b/cmd/autofix_test.go
@@ -2,7 +2,7 @@ package cmd
 
 import (
 	"bufio"
-	"bytes"
+	"fmt"
 	"os"
 	"testing"
 
@@ -25,7 +25,10 @@ func compareFiles(file1, file2 string) bool {
 
 	for s1.Scan() {
 		s2.Scan()
-		if !bytes.Equal(s1.Bytes(), s2.Bytes()) {
+		text1 := s1.Text()
+		text2 := s2.Text()
+		if text1 != text2 {
+			fmt.Printf("Files don't match here:\n%v\n%v\n\n", text1, text2)
 			return false
 		}
 	}
@@ -33,15 +36,13 @@ func compareFiles(file1, file2 string) bool {
 }
 
 func assertEqualWorkloads(assert *assert.Assertions, resource1, resource2 []Resource) {
-	file1 := "/tmp/dat1"
-	file2 := "/tmp/dat2"
-	err := WriteToFile(resource1[0], file1, false)
+	tmpfile1, err := WriteToTmpFile(resource1[0])
 	assert.Nil(err)
-	err = WriteToFile(resource2[0], file2, false)
+	defer os.Remove(tmpfile1)
+	tmpfile2, err := WriteToTmpFile(resource2[0])
 	assert.Nil(err)
-	assert.True(compareFiles(file1, file2))
-	os.Remove(file1)
-	os.Remove(file2)
+	defer os.Remove(tmpfile2)
+	assert.True(compareFiles(tmpfile1, tmpfile2))
 }
 
 func TestFixV1(t *testing.T) {
--- a/cmd/autofix_test.go
+++ b/cmd/autofix_test.go
@@ -1,12 +1,48 @@
 package cmd
 
 import (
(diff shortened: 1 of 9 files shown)


cmd/autofix_test.go
```diff
@@ -1,13 +1,49 @@
...
+	if err != nil {
+		return false
+	}
+
+	s1 := bufio.NewScanner(f1)
+	s2 := bufio.NewScanner(f2)
+
+	for s1.Scan() {
+		s2.Scan()
+		if !bytes.Equal(s1.Bytes(), s2.Bytes()) {
```
genevieveluyt: I'm sure this is more efficient but would it be more convenient for debugging to compare strings so that if the test fails it is easy to see the difference?

cmd/autofix_test.go
```diff
@@ -1,13 +1,49 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func assertEqualWorkloads(assert *assert.Assertions, resource1, resource2 []Resource) {
+	file1 := "/tmp/dat1"
```
genevieveluyt: Would it be beneficial to use io.ioutil.TempFile on the off chance that this file name is used by another program?

cmd/autofix_test.go
```diff
@@ -1,13 +1,49 @@
...
+	"bytes"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
-	"github.com/go-test/deep"
 )
 
+func compareFiles(file1, file2 string) bool {
```
genevieveluyt: +1 I will use this function in my PR :D

cmd/autofix_test.go
klautcomputing: Changed it to `.Text()` and added a `printf` for it so that you get notified about the difference.

cmd/test_util.go
```diff
@@ -112,3 +116,29 @@ func assertEqualYaml(fileToFix string, fileFixed string, auditFunc func(resource
 	assert.Nil(err)
 	assert.Equal(correctlyFixedResources[0], fixedResource)
 }
+
+// WriteToTmpFile writes a single resource to a tmpfile, you are responsible
+// for closing the file afterwards, that's why the function returns the file
```
genevieveluyt: Just nitpicking, do you mean deleting the file? It looks like it is closed on line 139 `err = tmpfile.Close()`

cmd/autofix_test.go
```diff
@@ -1,13 +1,50 @@
...
+	"fmt"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
-	"github.com/go-test/deep"
 )
 
+func compareFiles(file1, file2 string) bool {
```
chrisdonaldson: Is it worth being explicit that this only safely compares text files and will ignore EOL characters? (`\r?\n`)

cmd/test_util.go
```diff
@@ -112,3 +116,29 @@ func assertEqualYaml(fileToFix string, fileFixed string, auditFunc func(resource
...
+	if err != nil {
+		return "", err
+	}
+	_, err = tmpfile.Write(yaml)
+	if err != nil {
+		return "", err
+	}
+	err = tmpfile.Close()
+	if err != nil {
+		return "", err
```
chrisdonaldson: Why return an empty string on failed `.Close?` or write calls? Makes sense to me on open but in theory the filename might be useful to output on error by the caller?

cmd/autofix_test.go
klautcomputing: Why does this only safely compare text files? And why do you say it ignores EOL characters?

cmd/test_util.go
klautcomputing: There are only 2 cases in which an error can occur for a `file.Close()`, iirc. And correct me if I am wrong:
1. The file handle is `nil`. This would mean something super weird has happened because no other program should have the file handle because we just created it and this is single threaded so I don't see how it could have been `nil`.
1. The file has already be closed. If this is the case you already get an error that contains the filename in its error message.

So I cannot see the first one happen and for the second one you already get an informative error message.

chrisdonaldson: 👍

cmd/autofix_test.go
chrisdonaldson: `NewScanner` defaults to `ScanLines`. I'm still pretty new to this go jazz but `ScanLines` tries to drop carriage returns and newlines. The source seems to agree with my remembering: https://golang.org/src/bufio/scan.go?s=11799:11877#L345. 

If a non-text file is compared it would drop a bunch of probably meaningful, if misunderstood, bytes. It's a pretty unlikely situation where you end up with a false positive there in non-trivial cases I suppose.

Issue #127, opened 2018-10-23
Make sure types switches are exhaustive


klautcomputing:
Since this isn't Rust 😭 we need make sure our type switches do the thing correctly.

PR #147, opened 2019-01-22
Inherit Pod SecurityContext incase Container SecurityContext is not defined #138 

### LINKED ISSUE #138: RunAsNonRoot can be inherited from PodSecurityContext
Current check covers only Container SecurityContext, but `RunAsNonRoot`, `RunAsUser`, `RunAsGroup` and `SELinuxOptions` are all inherited from the PodSecurityContext unless they are defined explicitly container-wise. 

Please consider adding `PodSecurityContext` to the list of checked values.

Ref:
https://github.com/Shopify/kubeaudit/blob/87446f2a54a1c25e280697468fdfbf1b39d98517/cmd/runAsNonRoot.go#L10

nschhina:
## Description
Fixes #138, RANR check for nil CSC should be inherited from PSC if PSC is not nil
## Why
CSC is more specific to the container than PSC. So say if CSCNANR = false and PSCNANR = true, we'd use CSCNANR, the only case where PSC would make a difference is the case where CSCNANR is not defined(empty/nil) and PSCNANR = true, so in this case container inherits NANR from the pod, hence this shouldn't be an error.


cmd/k8sruntime_util.go
```diff
@@ -199,6 +199,16 @@ func getContainers(resource Resource) (container []ContainerV1) {
 	return container
 }
 
+// Get PodSpec from the PodV1 resource type to check for PSC
+
+func getPodSpecs(resource Resource) (podSpec PodSpecV1) {
+	switch kubeType := resource.(type) {
+	case *PodV1:
```
anthonyho007: do we need to use switch when there is only one resource type ?

cmd/runAsNonRoot.go
```diff
@@ -46,15 +47,65 @@ func checkRunAsNonRoot(container ContainerV1, result *Result) {
...
 	for _, container := range getContainers(resource) {
 		result, err := newResultFromResource(resource)
 		if err != nil {
 			log.Error(err)
 			return
 		}
 
-		checkRunAsNonRoot(container, result)
+		// check if Container Security Context is defined properly, else audit the Pod Security Context
+		if (container.SecurityContext == nil && podSpec.SecurityContext != nil) || ((container.SecurityContext != nil && container.SecurityContext.RunAsNonRoot == nil) && (podSpec.SecurityContext != nil && podSpec.SecurityContext.RunAsNonRoot != nil)) {
```
anthonyho007: I think it would be more clear if we just check
if isDefinedProperly(container) {
   checkRunAsRoot(container)
} else {
   ....
}

and ur conditional statements are a bit long, it would be nice if you can break it into multiple lines.

cmd/k8sruntime_util.go
nschhina: switch is used to by choice to maintain code style, also taking into consideration that we may add more resources to kubeaudit, also using type switch results in much cleaner code, if we're to use a if statement in this case, we'd be calling multiple functions on the resource to get the type. Also taking into account #127, I'd suggest that we use a switch for now

cmd/runAsNonRoot.go
nschhina: I agree with this suggestion, should create a new function to check PSC defined CSC not defined, would change this up

cmd/runAsNonRoot.go
```diff
@@ -46,15 +47,65 @@ func checkRunAsNonRoot(container ContainerV1, result *Result) {
 	return
 }
 
+// Checks the PSC for RANR
+
+func checkRunAsNonRootPSC(podSpec PodSpecV1, result *Result) {
```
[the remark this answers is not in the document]
nschhina: This is a good suggestion, I tried refactoring earlier but there's a small problem 
```SecurityContext *PodSecurityContext `json:"securityContext,omitempty" protobuf:"bytes,14,opt,name=securityContext"` ```
```SecurityContext *SecurityContext `json:"securityContext,omitempty" protobuf:"bytes,15,opt,name=securityContext"` ```

SecurityContext in ContainerV1 and PodSpecV1 object have different types, which would be troublesome if we pass SecurityContext to a single function from these two different objects.

File: cmd/k8sruntime_util.go (hunk 1 above)
Comment (anthonyho007) [reply cmd/k8sruntime_util.go]: ohh okay I see.

File: cmd/util.go [hunk 4]
```diff
@@ -360,3 +360,15 @@ func prettifyReason(reason string) string {
 	}
 	return reason
 }
+
+func isPSCDefinedCSCUndefined(podSpec PodSpecV1, container ContainerV1) bool {
```
anthonyho007: I thought u want to check if container SecurityContext is define properly or else use pod securityContext ? from #138 it says that SecurityContext is inherited from pod, so does it means that if container security context is defined then we could audit container else we just audit pod security context?

cmd/k8sruntime_util.go
anthonyho007: ohhh I see

klautcomputing: A type switch is the right thing here to do.

cmd/result.go
```diff
@@ -59,6 +59,11 @@ func createFields(res Result, occ Occurrence) (fields log.Fields) {
 		fields[k] = v
 	}
 	fields["Container"] = occ.container
+
+	if occ.id == ErrorRunAsNonRootFalse {
+		fields["Pod"] = occ.podHost
```
klautcomputing: Pod should only be set if there is actually something in the podHost variable because otherwise it shows up but is empty:
```
ERRO[0000] RunAsNonRoot is set to false (root user allowed), please set to true!  Container=fakeContainerRANR KubeType=pod Name=run_as_non_root_psc_false_csc_false Namespace=fakeDeploymentRANR Pod=
```

cmd/result.go
```diff
@@ -59,6 +59,11 @@ func createFields(res Result, occ Occurrence) (fields log.Fields) {
 		fields[k] = v
 	}
 	fields["Container"] = occ.container
+
```
klautcomputing: Now that we have errors that are not related to a container we should make sure to only set the container field if the container is actually present.

```
ERRO[0000] RunAsNonRoot is set to false (root user allowed), please set to true!  Container= KubeType=pod Name=run_as_non_root_false Namespace=fakeDeploymentRANR Pod=fakeRANR
```

cmd/runAsNonRoot.go
```diff
@@ -46,15 +47,65 @@ func checkRunAsNonRoot(container ContainerV1, result *Result) {
...
+	if reason := result.Labels["audit.kubernetes.io/allow-run-as-root"]; reason != "" {
+		if podSpec.SecurityContext == nil || podSpec.SecurityContext.RunAsNonRoot == nil || *podSpec.SecurityContext.RunAsNonRoot == false {
+			occ := Occurrence{
+				podHost:  podSpec.Hostname,
+				id:       ErrorRunAsNonRootFalseAllowed,
```
klautcomputing: I think we need different error cases for PSC and CSC

cmd/runAsNonRoot.go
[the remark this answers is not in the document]
nschhina: I see a couple of problems doing it the way you suggested- 1) We are also using podSpec.hostName and container.Name fields in respective functions so we'd have to add those two parameters + result+ the two boolean values 2) If you look again at the logic, we are not only using RunAsNonRoot, we are also checking if the SecurityContext is itself defined, which we could check in the auditRunAsNonRoot function but it'd get really ugly, it'd actually result in moving multiple nil checks to the parent function which imo is not a good idea. @klautcomputing  what do you think of this?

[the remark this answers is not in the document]
klautcomputing: I actually like this idea. It would also require to know whether the `context == nil` though

cmd/runAsNonRoot.go
```diff
@@ -46,15 +47,65 @@ func checkRunAsNonRoot(container ContainerV1, result *Result) {
...
+				message:  "Allowed setting RunAsNonRoot to false",
+				metadata: Metadata{"Reason": prettifyReason(reason)},
+			}
+			result.Occurrences = append(result.Occurrences, occ)
+		} else {
+			occ := Occurrence{
+				podHost:  podSpec.Hostname,
+				id:       ErrorMisconfiguredKubeauditAllow,
+				kind:     Warn,
+				message:  "Allowed setting RunAsNonRoot to false, but it is set to true",
```
klautcomputing: I think we want disjunct error messages, so that people know where the errors come from.

fixtures/run_as_non_root_psc_false_allowed_v1.yml
```diff
@@ -0,0 +1,17 @@
...
+    audit.kubernetes.io/allow-run-as-root: "Superuser privileges needed"    
+  namespace: fakeDeploymentRANR  
+spec:
+  securityContext:
+    runAsNonRoot: false
+  containers:
+  - name: fakeContainerRANR
+    resources: {}
+    securityContext: {}
+status: {}
```
klautcomputing: Please set your editor to include newline. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline
Same for all other files that don't have it

fixtures/run_as_non_root_psc_true_csc_false_v1.yml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
klautcomputing: This is actually okay. Because CSC wins over PSC.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#podsecuritycontext-v1-core

> Indicates that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. If unset or false, no such validation will be performed. May also be set in SecurityContext. If set in both SecurityContext and PodSecurityContext, the value specified in SecurityContext takes precedence.

fixtures/run_as_non_root_psc_true_csc_false_v1.yml
```diff
@@ -0,0 +1,15 @@
...
+kind: Pod
```
klautcomputing: That also means we will most likely want more than one test for this. Because we need to check how the behavior is with more than one container and what happens if some of the containers have it set, others don't, etc.

cmd/util.go
klautcomputing: ContainerSC wins over PodSC. But I left a longer explanation on one of the tests.

cmd/result.go
nschhina: this has been fixed. Great Suggestion!

cmd/result.go
nschhina: this has been fixed. Great Suggestion!

cmd/util.go
anthonyho007: it actually took me a while to fully understand what your function is try doing. I guess if you want to improve the readability you can change the function to check if isCSCDefined() eg), and u can just check for if container.ServurityConx != nil && contain.SecurityConx.RunAsNonRoot != nil {return true ; else false}

nschhina: We actually also need to check if PodSpec.SC != nil && PodSpec.SC.RANAR != nil, this is because by default we should point this error to be a Container SC error rather than it being a Pod SC error, this is solely because CSC takes precedence, but yes I have changed up the function a bit so it's looks cleaner 
```
func isCSCWellDefined(podSpec PodSpecV1, container ContainerV1) bool {
	if container.SecurityContext != nil && container.SecurityContext.RunAsNonRoot != nil {
		return true
	}
	if podSpec.SecurityContext == nil || podSpec.SecurityContext.RunAsNonRoot == nil {
		return true
	}
	return false
}
```
What do you think?

fixtures/run_as_non_root_psc_false_csc_nil_multiple_cont_v1.yml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
klautcomputing: I think this error:
```
fixtures/run_as_non_root_psc_false_csc_nil_multiple_cont_v1.yml
ERRO[0000] RunAsNonRoot is set to false (root user allowed) in Pods's Security Context and not set in Container's Security Context, please set to true!  KubeType=pod Name=run_as_non_root_psc_false_csc_nil_multiple Namespace=fakeDeploymentRANR Pod=multiple_containers
```
could be improved on, it should tell name the container that is violating RANR. 
Because otherwise if you add another container you just get the same error twice:
```
./kubeaudit nonroot -f fixtures/run_as_non_root_psc_false_csc_nil_multiple_cont_v1.yml + a new container
ERRO[0000] RunAsNonRoot is set to false (root user allowed) in Pods's Security Context and not set in Container's Security Context, please set to true!  KubeType=pod Name=run_as_non_root_psc_false_csc_nil_multiple Namespace=fakeDeploymentRANR Pod=multiple_containers
ERRO[0000] RunAsNonRoot is set to false (root user allowed) in Pods's Security Context and not set in Container's Security Context, please set to true!  KubeType=pod Name=run_as_non_root_psc_false_csc_nil_multiple Namespace=fakeDeploymentRANR Pod=multiple_containers
```

cmd/runAsNonRoot.go
```diff
@@ -29,32 +30,74 @@ func checkRunAsNonRoot(container ContainerV1, result *Result) {
 	} else if container.SecurityContext == nil || container.SecurityContext.RunAsNonRoot == nil {
 		occ := Occurrence{
 			container: container.Name,
-			id:        ErrorRunAsNonRootNil,
+			id:        ErrorRunAsNonRootPSCNilCSCNil,
 			kind:      Error,
-			message:   "RunAsNonRoot is not set, which results in root user being allowed!",
+			message:   "RunAsNonRoot is not set in Container/Pod's Security Context, which results in root user being allowed!",
```
klautcomputing: I am not 100% sure we want to talk about the pod in the error message here. But it's not something that will hold me back from merging. It just creates weird messages when there are more than 1 container.

cmd/util.go
```diff
@@ -360,3 +360,13 @@ func prettifyReason(reason string) string {
 	}
 	return reason
 }
+
+func isCSCWellDefined(podSpec PodSpecV1, container ContainerV1) bool {
+	if container.SecurityContext != nil && container.SecurityContext.RunAsNonRoot != nil {
```
klautcomputing: Does it make sense to check for `RunAsNonRoot != nil` because we already check it later?

nschhina: renamed the function to a better name of what it does, should be more clear now

cmd/runAsNonRoot.go
nschhina: Totally agree this could get confusing with multiple containers, removed Pod from the error message

fixtures/run_as_non_root_psc_false_csc_nil_multiple_cont_v1.yml
nschhina: added containerName field to the result occurrence, we would now be able to differentiate between two container errors

klautcomputing:
@jinankjain What do you think?

PR #153, opened 2019-02-07
Support for multi-container override label

### LINKED ISSUE #93: Allow labels don't support multiple containers
*Problem:* The current implementation implementation of labels doesn't allow to specify for which container the deviation is allowed. E.g. `kubeaudit.allow.capability.chown: "true"` has no information whether it is the `first` or `second` container if we have more than one container in a resource.
```
      containers:
      - name: frist
      - name: second
```
*Solution* Change labels to have a container that they refer to.

anthonyho007:
**Added support for the following override labels:**

- audit.kubernetes.io/allow-automount-service-account-token

- audit.kubernetes.io/pod/allow-read-only-root-filesystem-false
- container.audit.kubernetes.io/\<container-name\>/allow-read-only-root-filesystem-false

- audit.kubernetes.io/pod/allow-run-as-root
- container.audit.kubernetes.io/\<container-name\>/allow-run-as-root

- audit.kubernetes.io/pod/allow-privileged
- container.audit.kubernetes.io/\<container-name\>/allow-privileged

- audit.kubernetes.io/pod/allow-privilege-escalation
- container.audit.kubernetes.io/\<container-name\>/allow-privilege-escalation

- audit.kubernetes.io/pod/allow-capability-\<capability-name\>
- container.audit.kubernetes.io/\<container-name\>/allow-capability-<capability-name>

Fixes # (issue)
closes #93 

##### Type of change

- [x] Bug fix :bug:

##### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,25 +387,56 @@ spec:
     metadata:
       labels:
         apps: YourAppNameHere
-        audit.kubernetes.io/allow-privilege-escalation: "YourReasonForOverrideHere"
+        container.audit.kubernetes.io/<container-name>/allow-privilege-escalation: "YourReasonForOverrideHere"
 ```
 
 Any label with a non-nil reason string will prevent `kubeaudit` from throwing the corresponding error and issue a warning instead.
 Reasons matching `"true"` (not case sensitive) will be displayed as `Unspecified`.
 
-`kubeaudit` supports many labels:
-- [audit.kubernetes.io/allow-privilege-escalation](#allowpe_label)
-- [audit.kubernetes.io/allow-privileged](#priv_label)
-- [audit.kubernetes.io/allow-capability](#caps_label)
-- [audit.kubernetes.io/allow-run-as-root](#nonroot_label)
-- [audit.kubernetes.io/allow-automount-service-account-token](#sat_label)
-- [audit.kubernetes.io/allow-read-only-root-filesystem-false](#rootfs_label)
+`kubeaudit` can skip certain audits by applying override labels to containers. If you want skip an audit for a specific container inside a pod, you can add an container override label. For example, if you use `kubeaudit` to ignore `allow-run-as-root` check for container "MyContainer1", you can add the following label:
+
+```sh
+spec:
+  template:
+    metadata:
+      labels:
+        apps: YourAppNameHere
+        container.audit.kubernetes.io/MyContainer1/allow-run-as-root: "YourReasonForOverrideHere"
+```
+
+Similarly, you can have `kubeaudit` to skip a specific audit for all containers inside the pod by adding a pod override label. For example, if you use `kubeaudit` to ignore `allow-run-as-root` check for all containers inside the pod, you can add the following label:
+
+```sh
+spec:
+  template:
+    metadata:
+      labels:
+        apps: YourAppNameHere
+        audit.kubernetes.io/pod/allow-run-as-root: "YourReasonForOverrideHere"
+```
+
+`kubeaudit` supports many labels on pod or container level:
+- [audit.kubernetes.io/pod/allow-privilege-escalation](#allowpe_label)
+- [container.audit.kubernetes.io/\<container-name\>/allow-privilege-escalation](#allowpe_label)
+- [audit.kubernetes.io/pod/allow-privileged](#priv_label)
+- [container.audit.kubernetes.io/\<container-name\>/allow-privileged](#priv_label)
+- [audit.kubernetes.io/pod/allow-capability](#caps_label)
+- [container.audit.kubernetes.io/\<container-name\>/allow-capability](#caps_label)
+- [audit.kubernetes.io/pod/allow-run-as-root](#nonroot_label)
+- [container.audit.kubernetes.io/\<container-name\>/allow-run-as-root](#nonroot_label)
+- [audit.kubernetes.io/pod/allow-automount-service-account-token](#sat_label)
+- [audit.kubernetes.io/pod/allow-read-only-root-filesystem-false](#rootfs_label)
+- [container.audit.kubernetes.io/\<container-name\>/allow-read-only-root-filesystem-false](#rootfs_label)
 
 <a name="allowpe_label"/>
 
-### audit.kubernetes.io/allow-privilege-escalation
+### container.audit.kubernetes.io/\<container-name\>/allow-privilege-escalation
 
-Allows `allowPrivilegeEscalation` to be set to `true`.
+Allow `allowPrivilegeEscalation` to be set to `true` to a specific container.
+
+### audit.kubernetes.io/pod/allow-privilege-escalation
+
+Allows `allowPrivilegeEscalation` to be set to `true` to all the containers in a pod.
 
 ```sh
 kubeaudit.allow.privilegeEscalation: "Superuser privileges needed"
@@ -415,9 +446,13 @@ WARN[0000] Allowed setting AllowPrivilegeEscalation to true  Reason="Superuser p
 
 <a name="priv_label"/>
 
-### audit.kubernetes.io/allow-privileged
+### container.audit.kubernetes.io/\<container-name\>/allow-privileged
 
-Allows `privileged` to be set to `true`.
+Allow `privileged` to be set to `true` to a specific container.
+
+### audit.kubernetes.io/pod/allow-privileged
+
+Allows `privileged` to be set to `true` to all the containers in a pod.
 
 ```sh
 kubeaudit.allow.privileged: "Privileged execution required"
@@ -427,9 +462,13 @@ WARN[0000] Allowed setting privileged to true                Reason="Privileged 
 
 <a name="caps_label"/>
 
-### audit.kubernetes.io/allow-capability
+### container.audit.kubernetes.io/\<container-name\>/allow-capability
 
-Allows adding a capability or keeping one that would otherwise be dropped.
+Allows adding a capability or keeping one that would otherwise be dropped to a specific container.
+
+### audit.kubernetes.io/pod/allow-capability
+
+Allows adding a capability or keeping one that would otherwise be dropped to all the containers in a pod.
 
 ```sh
 kubeaudit.allow.capability.chown: "true"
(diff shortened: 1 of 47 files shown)


nschhina:
Can you update the checklist and mark the type of change, we are now enforcing everybody to use them

cmd/result.go
```diff
@@ -90,11 +90,18 @@ func shouldLog(err int) (members []string) {
...
 }
 
-func (res *Result) allowedCaps() (allowed map[CapabilityV1]string) {
+func (res *Result) allowedCaps(container ContainerV1) (allowed map[CapabilityV1]string) {
+
 	allowed = make(map[CapabilityV1]string)
 	for k, v := range res.Labels {
-		if strings.Contains(k, "audit.kubernetes.io/allow-capability-") {
-			capName := strings.Replace(strings.ToUpper(strings.TrimPrefix(k, "audit.kubernetes.io/allow-capability-")), "-", "_", -1)
+		if strings.Contains(k, "audit.kubernetes.io/pod/allow-capability-") {
```
klautcomputing: Could you please add a test for this code :) One that goes through both cases here.

README.md
```diff
@@ -387,25 +387,36 @@ spec:
...
-`kubeaudit` supports many labels:
-- [audit.kubernetes.io/allow-privilege-escalation](#allowpe_label)
-- [audit.kubernetes.io/allow-privileged](#priv_label)
-- [audit.kubernetes.io/allow-capability](#caps_label)
-- [audit.kubernetes.io/allow-run-as-root](#nonroot_label)
-- [audit.kubernetes.io/allow-automount-service-account-token](#sat_label)
-- [audit.kubernetes.io/allow-read-only-root-filesystem-false](#rootfs_label)
+Override label can be set on each containers or on a pod.
+
+`kubeaudit` supports many labels on pod or container level:
```
klautcomputing: I think we can make the doc a little less convoluted if we start with a longer explanation of how override labels can both be attached to a specific container or to a pod and then just describe one of them and then people can swap container-name for pod or the other way round.

fixtures/allow_privilege_escalation_true_single_allowed_multiple_containers_v1beta.yml
```diff
@@ -0,0 +1,20 @@
...
+        apps: fakeAllowPrivilegeEscalation
+        container.audit.kubernetes.io/fakeContainerAPE2/allow-privilege-escalation: "Superuser privileges needed"
+    spec:
+      containers:
+      - name: fakeContainerAPE
+        securityContext:
+            allowPrivilegeEscalation: true
+      - name: fakeContainerAPE2
+        securityContext:
+            allowPrivilegeEscalation: true
```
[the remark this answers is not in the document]
anthonyho007: Done

cmd/result.go
anthonyho007: Done

README.md
anthonyho007: I have added longer explanation on how pod level and container level override label work

anthonyho007:
@nschhina i have marked the checklist and type of changes

cmd/util.go
```diff
@@ -384,3 +384,36 @@ func shouldAuditCSC(podSpec PodSpecV1, container ContainerV1) bool {
 	}
 	return false
 }
+
```
nschhina: I don't think we need two different functions in `getContainerOverrideLabelReason` and `isDefinedContainerOverrideLabel` and same for Pods. Why don't we just return true or false from `getContainerOverrideLabelReason` currently defined and rename it to something like `containerOverrideLabelValid` ?

anthonyho007: nicee this is a good idea

anthonyho007: actually we need the overridelabelreason for the result/occurrence. i think its a bit neater to have two functions since i am using both in the code

nschhina: I don't see how it's useful for result/occurrence.. can you point out where? I also disagree with the neater part, `isDefinedContainerOverrideLabel` is literally `getContainerOverrideLabelReason` function where you return bool... if you are not using OverrideLabelReason anywhere outside checking if it exists and it matches I don't see the point of fetching the reason instead

anthonyho007: `else if container.SecurityContext == nil || container.SecurityContext.AllowPrivilegeEscalation == nil || *container.SecurityContext.AllowPrivilegeEscalation == true {
		occ := Occurrence{
			container: container.Name,
			id:        ErrorAllowPrivilegeEscalationTrueAllowed,
			kind:      Warn,
			message:   "Allowed AllowPrivilegeEscalation to be set as true",
			metadata:  Metadata{"Reason": prettifyReason(reason)},
		}`

anthonyho007: it is used in check-xxxx functions. I dont really see why are u against it, but if you really insist, i could just check reason != ""

nschhina: Ah it's used in Metadata, I guess we can leave it there then, I have a different suggestion in that case. Why don't we have one function and return string and bool from it? It would be much more elegant. Return reason as a string and bool dependent on whether label's defined or not?

README.md
```diff
@@ -387,25 +387,57 @@ spec:
     metadata:
       labels:
         apps: YourAppNameHere
-        audit.kubernetes.io/allow-privilege-escalation: "YourReasonForOverrideHere"
+        audit.kubernetes.io/pod/allow-privilege-escalation: "YourReasonForOverrideHere"
+        container.audit.kubernetes.io/<container-name>/allow-privileged: "YourReasonForOverrideHere"
```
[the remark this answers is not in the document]
anthonyho007: i have rewrote readme for override labels. it should be clearer now.

README.md
```diff
@@ -387,25 +387,57 @@ spec:
...
 
-`kubeaudit` supports many labels:
-- [audit.kubernetes.io/allow-privilege-escalation](#allowpe_label)
-- [audit.kubernetes.io/allow-privileged](#priv_label)
-- [audit.kubernetes.io/allow-capability](#caps_label)
-- [audit.kubernetes.io/allow-run-as-root](#nonroot_label)
-- [audit.kubernetes.io/allow-automount-service-account-token](#sat_label)
-- [audit.kubernetes.io/allow-read-only-root-filesystem-false](#rootfs_label)
+Override labels can be attached to a specific container or to a pod. Attaching an override label to a pod would allow certain audits to fail for all containers inside the pod. In another hand, attaching an override label to a specific container would allow certain audits to fail for that specific container. For example, if you `kubeaudit` to ignore `allow-run-as-root` check for container "MyContainer1", you can add the following label:
+
```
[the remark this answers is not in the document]
anthonyho007: i have rewrote readme for override labels. it should be clearer now.

cmd/util.go
anthonyho007: Done

README.md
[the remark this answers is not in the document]
nschhina: `allow-privileged` or `allow-privilege-escalation` for the label on line 390 ?

README.md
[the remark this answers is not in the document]
nschhina: "certain audit's to fail" doesn't sound right. What do you think of changing to something like "Allows the user to skip certain audits"

README.md
```diff
@@ -387,14 +387,13 @@ spec:
...
-        audit.kubernetes.io/pod/allow-privilege-escalation: "YourReasonForOverrideHere"
         container.audit.kubernetes.io/<container-name>/allow-privileged: "YourReasonForOverrideHere"
 ```
 
 Any label with a non-nil reason string will prevent `kubeaudit` from throwing the corresponding error and issue a warning instead.
 Reasons matching `"true"` (not case sensitive) will be displayed as `Unspecified`.
 
-Override labels can be attached to a specific container or to a pod. Attaching an override label to a pod would allow certain audits to fail for all containers inside the pod. In another hand, attaching an override label to a specific container would allow certain audits to fail for that specific container. For example, if you `kubeaudit` to ignore `allow-run-as-root` check for container "MyContainer1", you can add the following label:
+`kubeaudit` can allow certain audits to fail by applying override labels to containers. If you want to allow an audit to fail for a specific container inside a pod, you can add an container override label. For example, if you use `kubeaudit` to ignore `allow-run-as-root` check for container "MyContainer1", you can add the following label:
```
nschhina: Same here

cmd/util.go
```diff
@@ -385,26 +385,21 @@ func shouldAuditCSC(podSpec PodSpecV1, container ContainerV1) bool {
 	return false
 }
```
nschhina: Code changes look good, I would suggest we rename variable `exist` to `labelExists` for more verbosity

anthonyho007: Done

README.md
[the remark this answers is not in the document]
anthonyho007: Done

README.md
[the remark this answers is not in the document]
anthonyho007: Nice catch!!

nschhina:
LGTM, all my suggestions have been addressed. We can 🚢  after @klautcomputing gives his 👍

fixtures/capabilities_some_allowed_multi_containers_v2beta2.yml
```diff
@@ -0,0 +1,55 @@
+---
```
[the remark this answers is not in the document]
anthonyho007: I have rename them properly now

cmd/util.go
```diff
@@ -384,3 +384,31 @@ func shouldAuditCSC(podSpec PodSpecV1, container ContainerV1) bool {
...
+	if reason := result.Labels[podOverrideLabel]; reason != "" {
+		return true, reason
+	}
+	return false, ""
+}
+
+func isDefinedCapOverrideLabel(result *Result, container ContainerV1, capName string) bool {
+	capNameKey := strings.Replace(capName, "_", "-", -1)
+	containerKeyString := "container.audit.kubernetes.io/" + container.Name + "/allow-capability-" + capNameKey
+	if val, exist := result.Labels[containerKeyString]; exist && val != "" {
```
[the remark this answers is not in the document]
anthonyho007: I cant remember why i put that but i have changed it.

cmd/util.go
```diff
@@ -384,3 +384,31 @@ func shouldAuditCSC(podSpec PodSpecV1, container ContainerV1) bool {
...
+		return true
+	}
+
+	podKeyString := "audit.kubernetes.io/pod/allow-capability-" + capNameKey
+	return result.Labels[podKeyString] != ""
```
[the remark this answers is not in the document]
anthonyho007: Done

PR #158, opened 2019-02-13
add replacement for DSA to SA

### LINKED ISSUE #157: deprecatedServiceAccount not replaced with serviceAccountName
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

##### SUMMARY

deprecatedServiceAccount is not replaced with serviceAccountName in autofix

##### ENVIRONMENT
* Kubeaudit version: 0.4.1
* Kubeaudit install method: -

##### STEPS TO REPRODUCE

kubeaudit autofix -f fixtures/service_account_token_deprecated_v1.yml

##### EXPECTED RESULTS

deprecatedServiceAccount replaced with serviceAccountName in spec

##### ACTUAL RESULTS

deprecatedServiceAccount is set to nil(default)

nschhina:
##### Description

Kind words by Jon -

If serviceAccount is present, it should assume that serviceAccountName is also set and present because it's an alias (deprecated)

if there is no serviceAccount[Name] specified, it assumes `default` and uses the `default` SA token, which is not desirable in most multi-tenant circumstances, so automountServiceAccountToken: false should then be applied (edited) 
To use a non-default service account, simply set the spec.serviceAccountName field of a pod to the name of the service account you wish to use.
so realistically, if serviceAccount != default, then it should replace it with serviceAccountName instead
if it is default, remove it and add automount false

Fixes # (issue)
closes #157 
##### Type of change

- [x] Bug fix :bug:

Tests -
- [x] TestFixServiceAccountTokenDeprecatedV1

##### Checklist:

### DIFF
--- a/cmd/automountServiceAccountToken_fixes_test.go
+++ b/cmd/automountServiceAccountToken_fixes_test.go
@@ -11,6 +11,47 @@ func TestFixServiceAccountTokenDeprecatedV1(t *testing.T) {
 	}
 }
 
+func TestFixServiceAccountTokenDeprecatedV2(t *testing.T) {
+	assert, resources := FixTestSetupMultipleResources(t, "service_account_token_deprecated_multiple_resources_v1.yml", auditAutomountServiceAccountToken)
+	for index := range resources {
+		switch t := resources[index].(type) {
+		case *CronJobV1Beta1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.JobTemplate.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.JobTemplate.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DaemonSetV1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DaemonSetV1Beta1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DeploymentExtensionsV1Beta1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DeploymentV1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DeploymentV1Beta1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DeploymentV1Beta2:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *PodV1:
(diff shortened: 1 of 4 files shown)


klautcomputing:
It would be nice to have a regression test for this in case we break it by moving some code around.

nschhina:
Would add tests for all possible combinations next

cmd/test_util.go
```diff
@@ -34,6 +34,21 @@ func FixTestSetup(t *testing.T, file string, auditFunction func(Resource) []Resu
 	return assert, fixPotentialSecurityIssue(resource, result)
 }
 
+// FixTestSetupMultipleResources allows kubeaudit to be used programmatically instead of via the shell for multiple Resources. It is intended to be used for testing.
+func FixTestSetupMultipleResources(t *testing.T, file string, auditFunction func(Resource) []Result) (*assert.Assertions, []Resource) {
```
nschhina: @klautcomputing  How do you feel about using this test everywhere we were using `FixTestSetup`?

cmd/test_util.go
```diff
@@ -34,6 +34,21 @@ func FixTestSetup(t *testing.T, file string, auditFunction func(Resource) []Resu
...
+	var fixedResources []Resource
+	assert := assert.New(t)
+	file = filepath.Join(path, file)
+	resources, err := getKubeResourcesManifest(file)
+	assert.Nil(err)
+	results := getResults(resources, auditFunction)
+	for index := range resources {
+		resource := resources[index]
+		fixedResources = append(fixedResources, fixPotentialSecurityIssue(resource, results[index]))
```
anthonyho007: i wrote the same function too! Dont u want to fix all the issues in the file and return 1 resource ?

cmd/test_util.go
klautcomputing: Yeah sounds good to me. I would go for the shorter name though :)

cmd/automountServiceAccountToken_fixes_test.go
```diff
@@ -6,10 +6,52 @@ func TestFixServiceAccountTokenDeprecatedV1(t *testing.T) {
...
 }
 
+func TestFixServiceAccountTokenDeprecatedV2(t *testing.T) {
+	assert, resources := FixTestSetupMultipleResources(t, "service_account_token_deprecated_multiple_resources_v1.yml", auditAutomountServiceAccountToken)
+	for index := range resources {
+		switch t := resources[index].(type) {
+		case *CronJobV1Beta1:
+			assert.Equal("fakeDeprecatedServiceAccount", t.Spec.JobTemplate.Spec.Template.Spec.ServiceAccountName)
+			assert.Equal("", t.Spec.JobTemplate.Spec.Template.Spec.DeprecatedServiceAccount)
+		case *DaemonSetV1:
```
klautcomputing: I think you are missing a `DaemonSet` for `extensions/v1beta1` here.

cmd/k8sruntime_util.go
```diff
@@ -51,36 +51,47 @@ func setContainers(resource Resource, containers []ContainerV1) Resource {
 func disableDSA(resource Resource) Resource {
 	switch t := resource.(type) {
 	case *CronJobV1Beta1:
+		t.Spec.JobTemplate.Spec.Template.Spec.ServiceAccountName = t.Spec.JobTemplate.Spec.Template.Spec.DeprecatedServiceAccount
 		t.Spec.JobTemplate.Spec.Template.Spec.DeprecatedServiceAccount = ""
 		return t.DeepCopyObject()
 	case *DaemonSetV1:
+		t.Spec.Template.Spec.ServiceAccountName = t.Spec.Template.Spec.DeprecatedServiceAccount
```
klautcomputing: And here as well.

cmd/automountServiceAccountToken_fixes_test.go
nschhina: *DaemonSetV1Beta1 is the DaemonSet for extensions/v1beta 

`// DaemonSetV1Beta1 is a type alias for the v1beta1 version of the k8s extensions API.
type DaemonSetV1Beta1 = extensionsv1beta1.DaemonSet`

nschhina: appsv1/beta for DaemonSet is not a valid resource, that's why it was getting removed. I got rid of it now

PR #155, opened 2019-02-13
Autofix Network Policy

nschhina:
##### Description

Add autofix support for Network Policy auditing

For 🎩  try- 
```
kubeaudit autofix -f  fixtures/namespace_missing_default_deny_netpol.yml
```
Add default egress/ingress deny network policy to namespace

#159 to be fixed later

## Fixes (Autofix for Network Policies #156)

closes #156 

##### Type of change

- [x] Bug fix :bug:
- [x] New feature :sparkles:

##### How Has This Been Tested?

- [x]  TestAllResourcesFixV1
- [x] TestExtraResourcesFixV1
- [x] TestExtraResourcesIngressFixV1
- [x]  TestFixV1Beta2
- [x] TestFixV1Beta1
- [x]  TestFixV1

##### Checklist:

- [x] I have :tophat: my changes (A 🎩 specifically includes pulling down changes, setting them up, and manually testing the changed features and potential side effects to make sure nothing is broken)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes
- [x] The test coverage did not decrease

PR #164, opened 2019-02-20
Drop error to warn for Unsupported Workloads and add Regression Test for all Fixtures

### LINKED ISSUE #161: Unsupported Workloads should log WARN instead of ERRO with kubeaudit all audit
##### ISSUE TYPE
 - [x] Bug Report

##### SUMMARY

Unsupported Workloads should log WARN instead of ERRO with kubeaudit all audit. Currently we log this as Error, instead this should be logged as a warning.

##### ENVIRONMENT
* Kubeaudit version: 0.4.1
* Kubeaudit install method: 

##### STEPS TO REPRODUCE

run
```
kubeaudit all -f fixtures/unknown_type_v1.yml
```

##### EXPECTED RESULTS

Should log all result instances as warnings

##### ACTUAL RESULTS

They are logged as ERROR

### LINKED ISSUE #162: Add Regression Test to test.sh
##### ISSUE TYPE
 - [x] Bug Report

##### SUMMARY

Test if autofix handles clears all errors for all files in fixtures/
##### ENVIRONMENT
* Kubeaudit version: 0.4.1
* Kubeaudit install method: -

##### STEPS TO REPRODUCE

Add Test for this

nschhina:
MERGE AFTER #155 and #158 

##### Description

Drop error to warn for Unsupported Workloads and add Regression Test for all Fixtures.

Fixes # (issue)
closes #162, closes #161 

##### Type of change

- [x] Bug fix :bug:

- [x] All Present Tests Pass along with the added regression test

##### Checklist:


nschhina:
Not sure how we should handle Double-Container Test here -https://github.com/Shopify/kubeaudit/pull/164/files#diff-599a5c6f02518607fa2004a102063847R9 and the manifest here - https://github.com/Shopify/kubeaudit/blob/325b0b5214367ff247894c8969987f3552c29114/fixtures/double-name_v1.yml. 

The regression test I've added to test.sh autofixes all files in fixtures/* and then audits all the autofixed files, and the test returns 1 if any "ERROR" logs still exist in all those fixtures/* . We don't have any autofix for duplicate container name and I don't think we should, this error should be handled manually. I suggest we create a different directory "manual-fixtures" for such cases. Open to suggestions right now.

klautcomputing:
I will review this after we have merged the two other branches because otherwise I am looking at a lot of code that I already looked at and will just repeat myself.

Issue #165, opened 2019-02-21
Code Coverage should be better than 80%


nschhina:
##### ISSUE TYPE
 - [x] Feature Idea

##### SUMMARY

Currently Code Coverage is about 76% on the dropErrors branch, we should hit 80% minimum by next release or the release after. 

##### ENVIRONMENT
* Kubeaudit version: 0.4.1
* Kubeaudit install method: -

### FEATURE IDEA

- [x] If the maintainers agree with the feature as described here, I intend to submit a Pull Request myself.<sup>1</sup>

**Proposal: Would open up an PR to increase coverage** 

<sup><small>1</small></sup> <sub>This is the quickest way to get a new feature! We reserve the right to close feature requests, even ones we like, if the proposer does not intend to contribute to the feature and it doesn't fit in our current roadmap.</sub>

PR #166, opened 2019-02-21
Increase Code Coverage

nschhina:
##### Description
Code Coverage is now 81%, barring error messages, warnings and clientset everything else is pretty much covered.

Fixes # (issue)
#165  

##### Checklist:


cmd/automountServiceAccountToken_fixes.go
```diff
@@ -1,9 +1,6 @@
 package cmd
 
 func fixServiceAccountToken(result *Result, resource Resource) Resource {
-	if labelExists, _ := getPodOverrideLabelReason(result, "allow-automount-service-account-token"); labelExists {
```
klautcomputing: This `if` always returns false, so it can be deleted.

fixtures/capSetConfig.yaml
```diff
@@ -0,0 +1,15 @@
+capabilitiesToBeDropped:
```
klautcomputing: Is this missing one cap on purpose? If so please document that behavior, if not please add it.

PR #172, opened 2019-02-26
Fix default SA name bug

### LINKED ISSUE #168: KubeAudit should raise an error if ServiceAccountName is set to "default"
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

# BUG REPORT

##### SUMMARY
if there is no serviceAccount[Name] specified, it assumes `default` and uses the `default` SA token, which is not desirable in most multi-tenant circumstances, so automountServiceAccountToken: false should then be applied. so realistically, if serviceAccount != default, then it should replace it with serviceAccountName instead. if it is default, remove it and add automount false

##### ENVIRONMENT
* Kubeaudit version: master
* Kubeaudit install method: master 

##### STEPS TO REPRODUCE

- include serviceAccountName: "default" in the pod spec
- user kubeaudit to audit the yml file

##### EXPECTED RESULTS

Kubeaudit should raise an error

##### ACTUAL RESULTS

not error is raised

nschhina:
##### Description

Ref #168, When `automountServiceAccountToken` is set to true and `serviceAccountName` is set to be `default` kubeaudit didn't throw an error as it should, this issue has been addressed in this PR.

Fixes # (issue)
closes #168 
##### Type of change

- [x] Bug fix :bug:

- [x] TestServiceAccountTokenTrueAndDefaultNameV1

##### Checklist:

### DIFF
--- a/cmd/automountServiceAccountToken.go
+++ b/cmd/automountServiceAccountToken.go
@@ -38,7 +38,7 @@ func checkAutomountServiceAccountToken(result *Result) {
 		return
 	}
 
-	if result.Token != nil && *result.Token && result.SA == "" {
+	if result.Token != nil && *result.Token && (result.SA == "" || result.SA == "default") {
 		// automountServiceAccountToken = true, and serviceAccountName is blank (default: default)
 		occ := Occurrence{
 			id:      ErrorAutomountServiceAccountTokenTrueAndNoName,
--- a/cmd/automountServiceAccountToken.go
+++ b/cmd/automountServiceAccountToken.go
@@ -38,7 +38,8 @@ func checkAutomountServiceAccountToken(result *Result) {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/automountServiceAccountToken.go
```diff
@@ -38,7 +38,7 @@ func checkAutomountServiceAccountToken(result *Result) {
 		return
 	}
 
-	if result.Token != nil && *result.Token && result.SA == "" {
+	if result.Token != nil && *result.Token && (result.SA == "" || result.SA == "default") {
```
catherinejones: This is a very minor nitpick but could you split this if statement into multiple lines? Ideally put the `(result.SA == "" || result.SA == "default")` on its own line, so it's more readable?

nschhina: Done!

PR #173, opened 2019-02-26
Use Go 1.12

### LINKED ISSUE #170: Support and Test Go 1.12
# BUG REPORT

##### SUMMARY

Go 1.12 is now released and `kubeaudit` should be compatible with it, so we need to replace Go 1.10 with Go 1.12 for our CI environment.

##### ENVIRONMENT
* Kubeaudit version: 0.4.0
* Kubeaudit install method: -

nschhina:
##### Description
Ref #170 
Go 1.12 is now released and `kubeaudit` should be compatible with it, so we need to replace Go 1.10 with Go 1.12 for our CI environment.

Fixes # (issue)
closes #170

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - "1.10.x"
   - "1.11.x"
+  - "1.12.x"
 
 script:
   - make setup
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - "1.10.x"
```
chrisdonaldson: Why remove the 1.10.x?

nschhina: Go standards suggest that it's users responsibility to make sure that his goversion is up to date. Leaving 1.10 in there would only increase our build time. Moreover plan is to introduce `go mod` soon, which isn't backward compatible with 1.10

chrisdonaldson: :+1

PR #174, opened 2019-02-27
Replace Dep with Go Modules 

### LINKED ISSUE #171: Support Go Modules
##### SUMMARY

Ref #170  we will be moving from Go 1.10,1.11 to Go 1.11,1.12 and `go mod` is supported 1.11 onwards. So we should add support for this

##### ENVIRONMENT
* Kubeaudit version: 0.4.0
* Kubeaudit install method: -

nschhina:
##### Description
Ref #171 we will be moving from Go 1.10,1.11 to Go 1.11,1.12 and `go mod` is supported 1.11 onwards. This PR adds the support for `go mod`

Fixes # (issue)
closes #171 
##### Type of change

- [x] New feature :sparkles:

##### Checklist:


Makefile
```diff
@@ -44,8 +45,7 @@ check_version:
 	./$(BINARY_NAME) version
 
 setup:
-	$(GOCMD) get -u github.com/golang/dep/cmd/dep
-	$(GODEP) ensure
+	go mod download
```
klautcomputing: I think we should add a `go mod tidy` somewhere

.gitignore
```diff
@@ -4,7 +4,6 @@
 /.glide
 /.dev
 .DS_Store
-/vendor
```
klautcomputing: since you don't commit any thing in vendor we don't need to include it, do we?

nschhina: Correct, vendor is useless now, we'll not have any vendor files anymore. That's why I excluded it from gitignore

Makefile
nschhina: Done!

.gitignore
klautcomputing: Okay this just means we all need to delete `vendor` going forward 👍

nschhina: We also need to `export GO111MODULE=on` if we are working on kubeaudit in $GOPATH

PR #227, opened 2019-09-23
Replace Shopify/yaml with gopkg.in/yaml.v3

### LINKED ISSUE #223: Replace our fork of go-yaml with go-yaml.v3
##### ISSUE TYPE
 - [ ] Bug Report
 - [ ] Feature Idea
 - [x] Maintainability Improvement

We are using a fork of go-yaml (Shopify/yaml) because go-yaml.v2 did not support comments. go-yaml.v3 is out now (https://gopkg.in/yaml.v3) and supports comments, so we should migrate over to using that instead so we get all the latest bug fixes and features and can stop maintaining the fork.

genevieveluyt:
##### Description

Now that goyaml v3 is out and supports comments, we can use that instead of the Shopify fork of goyaml v2.

~~There is one test that fails because of https://github.com/go-yaml/yaml/issues/508. Once that is fixed, we need to update the goyaml v3 dependency and then hopefully all the tests will pass!~~

Fixes #223 

##### Type of change

- [ ] Bug fix :bug:
- [ ] New feature :sparkles:
- [x] Code improvement

- [x] Run the test script `test.sh`

##### Checklist:

### DIFF
--- a/cmd/autofix_test.go
+++ b/cmd/autofix_test.go
@@ -6,8 +6,8 @@ import (
 	"os"
 	"testing"
 
-	"github.com/Shopify/yaml"
 	"github.com/stretchr/testify/assert"
+	yaml "gopkg.in/yaml.v3"
 )
 
 func TestFixV1(t *testing.T) {
@@ -112,7 +112,6 @@ func TestFixV1Beta2(t *testing.T) {
 	autofix(nil, nil)
 
 	assert.True(compareTextFiles(expectedFilename, tmpFilename))
-
 }
 
 func TestInvalidManifests(t *testing.T) {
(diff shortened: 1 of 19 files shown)


cmd/autofix_test.go
```diff
@@ -216,422 +215,597 @@ func TestSingleUnsupportedResource(t *testing.T) {
...
+			{Kind: yaml.ScalarNode, Tag: strTag, Value: "V", LineComment: "Bye"},
+		}},
+		&yaml.Node{Kind: yaml.MappingNode, Tag: mapTag, Content: []*yaml.Node{
+			{Kind: yaml.ScalarNode, Tag: strTag, Value: "K"},
+			{Kind: yaml.ScalarNode, Tag: strTag, Value: "V2"},
+		}},
+		&yaml.Node{Kind: yaml.MappingNode, Tag: mapTag, Content: []*yaml.Node{
+			{Kind: yaml.ScalarNode, Tag: strTag, Value: "K", HeadComment: "Hi"},
+			{Kind: yaml.ScalarNode, Tag: strTag, Value: "V2", LineComment: "Bye"},
+		}},
```
actgardner: nit: because this is kind of a big PR, I would recommend trying to keep as many things constant as possible to make it easy for reviewers. In this test case both the structure and the content of the nodes change, which makes things a little more complicated - especially because I'm new to the codebase. If we want to test both comment types and we didn't support them before it might make sense to split out multiple smaller commits that reviewers can consider individually like: "move to yaml.v3 nodes", "add HeadComment to test cases"? Just a suggestion for future PRs.

fixtures/preserve_comments-fixed_v2.yml
```diff
@@ -35,10 +35,7 @@ spec:
   containers:
   - name: test-container
     image: k8s.gcr.io/busybox
-    command:
-    - /bin/sh
-    - -c
-    - ls /etc/config/
+    command: ["/bin/sh", "-c", "ls /etc/config/"]
```
actgardner: These changes are kind of surprising, since the library is supposed to preserve the structure of the YAML file? Is this caused by the move to yaml.v3?

genevieveluyt: Actually this output matches the formatting of the input file https://github.com/Shopify/kubeaudit/blob/master/fixtures/preserve_comments_v2.yml#L35
Our fork used default formatting for all data types so all arrays got marshaled into the multi-line dashed formatting. It wasn't great but it wasn't technically incorrect as far as yaml parsers could tell.

cmd/autofix_test.go
genevieveluyt: Yes, that makes sense! We did support them before, they just looked different. In the old format, aka what the fork uses, it would have looked like this:
```
yaml.MapSlice{
	{Comment: "Hi"},
	{Key: "K", Value: "V2", Comment: "Bye"},
},
```
but I agree that I should have done a direct conversion and kept the test contents the same. I'll keep it in mind for next time!

fixtures/autofix-all-resources-fixed_v1.yml
```diff
@@ -189,7 +189,8 @@ spec:
     spec:
       # Trust me the following line is longer than 80 characters
       containers:
-      - image: gcr.io/docker-images-directory/apps/production/fake-image-app:<%= current_sha %>
+      - image: gcr.io/docker-images-directory/apps/production/fake-image-app:<%= current_sha
```
klautcomputing: Is there a way to not do the 80 (or whatever it is) character limit?

fixtures/autofix-namespace-missing-default-deny-netpol.yml
```diff
@@ -21,5 +21,5 @@ spec:
         podSelector:           # chooses pods with type=monitoring
           matchLabels:
             type: monitoring
-    - ports: # This is comment 1
-      - port: 3452 #This is comment 2
+    - ports:
```
klautcomputing: What happened to this and the other comments?

fixtures/preserve_comments-fixed_v2.yml
klautcomputing: I like the one liner way more :)

fixtures/autofix-all-resources-fixed_v1.yml
genevieveluyt: I doesn't look like it's supported right now but I could open a feature request

fixtures/autofix-namespace-missing-default-deny-netpol.yml
genevieveluyt: These caused tests to fail because of the bug in goyaml but since they're covered by the PreserveComments test, I took them out to make sure nothing else was causing the tests to fail. I can put them back but I think the PreserveComments test is thorough enough. What do you think?

Issue #233, opened 2019-10-08
error with custom resources


vvuibert:
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

# BUG REPORT

##### SUMMARY
we use [statefulservice](https://github.com/Shopify/cloudbuddies/blob/master/apis/shopify/v1/statefulservice.go) which is a custom resource and `kubeaudit` throughs an error instead of a warning since it's not supported

##### ENVIRONMENT
* Kubeaudit version: 0.0.0
* Kubeaudit install method: DIY-BUILD

##### STEPS TO REPRODUCE
`kubeaudit autofix -f config/deploy/staging/kafka.yml`
https://github.com/Shopify/kafka-infrastructure/blob/master/config/deploy/staging/kafka.yml

##### EXPECTED RESULTS

WARN Skipping unsupported resource type Statefulservice

##### ACTUAL RESULTS
```
kubeaudit autofix -f config/deploy/staging/kafka.yml
WARN[0000] Skipping unsupported resource type /v1, Kind=Service
WARN[0000] Skipping unsupported resource type /v1, Kind=Service
WARN[0000] Skipping unsupported resource type policy/v1beta1, Kind=PodDisruptionBudget
ERRO[0000] File is not a valid Kubernetes manifest: no kind "Statefulservice" is registered for version "stable.shopify.io/v1" in scheme "pkg/runtime/scheme.go:101"
```

##### ADDITIONAL INFORMATION

# FEATURE IDEA

- [ ] If the maintainers agree with the feature as described here, I intend to submit a Pull Request myself.<sup>1</sup>

**Proposal:** 

<sup><small>1</small></sup> <sub>This is the quickest way to get a new feature! We reserve the right to close feature requests, even ones we like, if the proposer does not intend to contribute to the feature and it doesn't fit in our current roadmap.</sub>

genevieveluyt:
This should be fixed by #243. Please re-open if the problem remains

PR #234, opened 2019-10-10
Kubeaudit package base PR

### LINKED ISSUE #239: Kubeaudit Package
##### ISSUE TYPE
 - [ ] Bug Report
 - [x] Feature Idea

# FEATURE IDEA

**Proposal:** 
Make kubeaudit importable as a package rather than only a CLI.

genevieveluyt:
##### Description

Makes it so you can use kubeaudit as a package and as a CLI!

Closes #239 and probably also #235 and #233 

This is the "base PR" that includes only the general scaffolding and a single auditor.

Things that still need to happen:
- [ ] Open separate PRs for each auditor and merge them into this one
- [ ] Add all the auditors to `auditors/all/all.go`
- [ ] Add the CLI and demo code
- [ ] Update the readme

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [ ] `make test`

##### Checklist:


internal/k8s/client.go
```diff
@@ -0,0 +1,166 @@
...
+	return netPols
+}
+
+func GetNamespaces(clientset kubernetes.Interface, namespace string) *k8stypes.NamespaceListV1 {
+	namespaceClient := clientset.CoreV1().Namespaces()
+	listOptions := k8stypes.ListOptionsV1{}
+
+	if namespace != "" {
+		// Select only the specified namespace
+		// TODO this doesn't work??
```
catherinejones: Concerning TODO here!

internal/k8s/client.go
```diff
@@ -0,0 +1,166 @@
...
+
+var DefaultClient = K8sClient{}
+
+// Client abstracts the API to allow testing.
+type Client interface {
+	InClusterConfig() (*rest.Config, error)
+}
+
+// K8sClient wraps kubernetes client-go so it can be mocked.
+type K8sClient struct{}
```
catherinejones: Does this structure need to be exported?

internal/k8s/client.go
```diff
@@ -0,0 +1,166 @@
...
+
+func NewKubeClientLocal(configPath string) (*kubernetes.Clientset, error) {
+	if _, err := os.Stat(configPath); err != nil {
+		return nil, ErrNoReadableKubeConfig
+	}
+	config, err := clientcmd.BuildConfigFromFlags("", configPath)
+	if err != nil {
+		return nil, err
+	}
+	kube, err := kubernetes.NewForConfig(config)
```
catherinejones: Is there a reason we're not using the KUBECONFIG environment variable to determine the configuration to use? We do this here: https://github.com/Shopify/vouch4cluster/blob/master/cmd/kube.go#L31

internal/k8s/client.go
```diff
@@ -0,0 +1,166 @@
...
+func IsRunningInCluster(client Client) bool {
+	config, err := client.InClusterConfig()
+	fmt.Println("Config ", config)
+	return err == nil
+}
+
+func GetAllResources(clientset kubernetes.Interface, namespace string) []k8stypes.Resource {
+	var resources []k8stypes.Resource
+
+	for _, resource := range GetDaemonSets(clientset, namespace).Items {
```
catherinejones: Could you make a type out of the `Get*` functions that have the same signature? 

```golang
type resourceGetter func(clientset kubernetes.Interface, namespace string) *k8stypes.DaemonSetListV1
```

And then replace this with:

```golang
for _, resourceGetter := range []resourceGetter{
    GetDaemonSets,
    GetDeployments,
} {
       for _, resource := range resourceGetter(clients, namespace).Items {
            resources = append(resources, resource.DeepCopyObject())
       }
}
```

internal/k8s/client_test.go
```diff
@@ -0,0 +1,122 @@
...
+			resource := template.DeepCopyObject()
+			setNamespace(resource, namespace)
+			resources = append(resources, resource)
+		}
+	}
+
+	clientset := fakeclientset.NewSimpleClientset(resources...)
+	assert.Len(t, GetAllResources(clientset, ""), len(resourceTemplates)*len(namespaces))
+
+	// TODO uncomment this when GetNamespaces() is fixed
```
catherinejones: Should this be fixed before this PR is merged?

internal/k8s/runtime_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func TestEncodeDecode(t *testing.T) {
+	assert := assert.New(t)
+
+	deployment := k8stypes.NewDeployment()
+	deployment.ObjectMeta = k8stypes.ObjectMetaV1{Namespace: "foo"}
+	deployment.Spec.Template.Spec.Containers = []k8stypes.ContainerV1{{Name: "bar"}}
+
+	expectedManifest := `apiVersion: apps/v1
+kind: Deployment
```
Nilay-Shah: Nitpick: can this block be indented twice so it's properly set with everything else?

internal/k8s/client.go
genevieveluyt: Yes although I'm convinced it never worked so unless someone knows an easy fix or can confirm it actually does work I'll open a separate issue for it

auditors/apparmor/apparmor.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+const (
+	// AppArmorAnnotationMissing occurs when the apparmor annotation is missing
+	AppArmorAnnotationMissing = "AppArmorAnnotationMissing"
+	// AppArmorDisabled occurs when the apparmor annotation is set to a bad value
+	AppArmorDisabled = "AppArmorDisabled"
+)
+
+// As of Oct 1, 2018 these constants are not in the K8s API package, but once they are they should be replaced
```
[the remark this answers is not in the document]
genevieveluyt: I updated this when I started working on the refactor, but since that was forever ago I'll update it again 😄

fix.go
```diff
@@ -0,0 +1,135 @@
...
+
+	fixedresourceBytes, err = cleanupManifest(origResourceBytes, fixedresourceBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return fixedresourceBytes, nil
+}
+
+// TODO do this better??
```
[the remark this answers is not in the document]
genevieveluyt: I didn't change this code but it seems pretty finicky to me. As in, I think in some cases it produces invalid yaml. What do you think? If it's a reasonable concern and not an easy fix I'll open an issue for it.

internal/k8s/runtime.go
```diff
@@ -0,0 +1,172 @@
...
+
+// NewTrue returns a pointer to a boolean variable set to true
+func NewTrue() *bool {
+	b := true
+	return &b
+}
+
+// NewFalse returns a pointer to a boolean variable set to false
+func NewFalse() *bool {
+	return new(bool)
```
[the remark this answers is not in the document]
genevieveluyt: This code existed already so I'm not sure what the original intent was behind the different implementation. I think using `new` is preferred since it initializes an object and returns a pointer to it, which is exactly what we want. But in the case of `true` I don't think there is a way to do it without assigning to a separate variable first.

These functions are used by a bunch of the other auditors to modify a container's security context because all the fields expect pointers
https://godoc.org/k8s.io/api/core/v1#SecurityContext

internal/k8s/runtime.go
```diff
@@ -0,0 +1,172 @@
...
+		return &kubeType.Template
+	case *k8stypes.ReplicationControllerV1:
+		return kubeType.Spec.Template
+	case *k8stypes.StatefulSetV1:
+		return &kubeType.Spec.Template
+	case *k8stypes.StatefulSetV1Beta1:
+		return &kubeType.Spec.Template
+	}
+
+	// Pod, Namespace
```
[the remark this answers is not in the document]
genevieveluyt: No, we have a list of supported resource types so I just wanted to be explicit that we are returning `nil` for those resources and haven't just forgotten about them. Would it be more clear to have
```
case *k8stypes.PodV1, *k8stypes.NamespaceV1:
        return nil
```
?

internal/k8s/client.go
genevieveluyt: I think when I tried to do this I got stuck on the fact that `Items` is a struct field and not a function call so what would the return value of the shared function signature be?

internal/k8s/client_test.go
genevieveluyt: I don't think so... but I suppose it depends if we think fixing GetNamespaces should be part of this PR

internal/k8s/client.go
catherinejones: The returned value would be whatever structure contains `Items`!

internal/k8s/client_test.go
catherinejones: Based on: https://github.com/Shopify/kubeaudit/pull/234#discussion_r366515799

if it never worked, I don't think we should worry about it now?

internal/k8s/client.go
catherinejones: Could we open an issue to fix this? (see also https://github.com/Shopify/kubeaudit/pull/234#discussion_r366559035)

internal/k8s/runtime.go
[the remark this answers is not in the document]
Nilay-Shah: Yeah, I'd personally find that much more clearer!

internal/k8s/runtime.go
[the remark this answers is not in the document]
Nilay-Shah: TIL. Thanks for the explanation!

fix.go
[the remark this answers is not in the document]
Nilay-Shah: Oh true! 
How do you know it produces bad YAML? I think we should definitely address that (probably not in the scope of this PR, though)

internal/k8s/client.go
genevieveluyt: Hmm but it's a different structure for each resource type?
DaemonSetList https://godoc.org/k8s.io/api/apps/v1#DaemonSetList
Deployment https://godoc.org/k8s.io/api/apps/v1#DeploymentList

fix.go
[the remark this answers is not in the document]
genevieveluyt: Because it just remove lines of YAML
eg.
```
kind: Pod
metadata:
    creationTimestamp: null
spec:
    containers:
    - name: AAcontainer
```
would turn into
```
kind: Pod
metadata:
spec:
    containers:
    - name: AAcontainer
```
(where `creationTimestamp: null` is removed)
which maybe is valid yaml but I don't think it's a valid kubernetes config. I think the correct fix for this is upstream. I had a quick look and they do have it set to only marshal that field if it's not empty, but because the type is a date object it can't tell when it's empty. There are other cases where they use a pointer to a date object and setting that to null won't produce the field in the yaml, which is what we want. Not sure why some dates are pointers and others aren't...

internal/k8s/client_test.go
genevieveluyt: I think that's fair. Do you think I should take the test out then or leave it as a todo?

fix.go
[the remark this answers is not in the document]
Nilay-Shah: Ahhh I see. I think it's worth creating an issue on our end to keep track of things and explore this farther as well? Thoughts?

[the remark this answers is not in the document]
genevieveluyt: Done https://github.com/Shopify/kubeaudit/issues/240

internal/k8s/runtime_test.go
genevieveluyt: Hmm It's a string comparison so we would need to add indents to the programatically produced manifest for them to still be equal, or like take out the indents before comparing. I could move the variable out of the function so the variable name is at least indented the same level as the rest of the file? What do you think?

Nilay-Shah: Ah right, makes sense. Can you move the `apiVersion: apps/v1` onto a newline so it's indented the same level as the rest of the expected manifest?

genevieveluyt: We would have to prepend a newline to the programatically produced output but yes we could do that. Do you think it makes sense to have the "expected output" in the test be different from the "actual expected output" though? As in, we don't want the manifest to start with a newline (because that would be an empty line at the top of the file), but we would be setting that as the expected output in the test for formatting purposes.

genevieveluyt: What do you think about moving it into a yaml file?

internal/k8s/client.go
genevieveluyt: https://github.com/Shopify/kubeaudit/issues/241

internal/k8s/client.go
genevieveluyt: I think kubeaudit just never supported using that environment variable. From the readme:

> Local config mode: If kubeaudit is not running in a container, `kubeaudit cmd` will audit the resources specified by your local kubeconfig (`$HOME/.kube/config`) file. You can force kubeaudit to use a specific local config file with the switch `-c/--kubeconfig /config/path`

ref: https://github.com/Shopify/kubeaudit#general-instructions

I'm not against changing it, but do you think it should be part of this PR?

fix.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Fix all the resources
+	for _, result := range results {
+		for _, auditResult := range result.GetAuditResults() {
+			newResources = append(newResources, auditResult.Fix(result.GetResource().Object())...)
+		}
+	}
+
+	// Convert all the resources to bytes
+	for _, result := range results {
```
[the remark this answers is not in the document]
genevieveluyt: For sure, I mostly just split it up to make the code easy to read

internal/k8s/runtime_test.go
Nilay-Shah: I'm not too familiar with idiomatic Golang when it comes to having fixtures, but I think having a fixtures directory to keep this new yaml file (along with any other ones we have now and/or in the future) will be beneficial?

fix.go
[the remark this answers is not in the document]
genevieveluyt: Were you thinking something like
```
func fix(results []Result) ([]byte, error) {
	var resourceBytes [][]byte
	var newResourceBytes [][]byte

	if len(results) == 0 {
		return []byte{}, nil
	}

		for _, auditResult := range result.GetAuditResults() {
			for _, newResource := range auditResult.Fix(result.GetResource().Object()) {
				fixedresourceBytes, err := resourceToBytes(newResource, nil)
				if err != nil {
					return nil, err
				}
				newResourceBytes = append(newResourceBytes, fixedresourceBytes)
			}
		}

		if result.GetResource().Object() == nil {
			resourceBytes = append(resourceBytes, result.GetResource().Bytes())
			continue
		}

		fixedresourceBytes, err := resourceToBytes(result.GetResource().Object(), result.GetResource().Bytes())
		if err != nil {
			return nil, err
		}

		resourceBytes = append(resourceBytes, fixedresourceBytes)
	}

	resourceBytes = append(resourceBytes, newResourceBytes...)
	fixedManifest := bytes.Join(resourceBytes, []byte("---"))

	return fixedManifest, nil
}
```

internal/k8s/runtime_test.go
genevieveluyt: I'm honestly not too sure either. The old version (before the refactor) had a single fixtures directory and I personally thought that was kind of a mess, so I created a fixtures directory for each auditor. I think having a fixtures directory inside any directory that has test files is fine though. What do you think?

Nilay-Shah: Yeah, this makes sense to me!

fix.go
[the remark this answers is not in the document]
Nilay-Shah: Yeah! This seems more performant than what you had previously, and I think the sacrifice in readability is worth it? What do you think?

fix.go
```diff
@@ -0,0 +1,135 @@
...
+		outputBytes = append(outputBytes, fixedresourceBytes)
+	}
+
+	fixedManifest := bytes.Join(outputBytes, []byte("---"))
+
+	return fixedManifest, nil
+}
+
+func resourceToBytes(fixedResource k8stypes.Resource, origResourceBytes []byte) ([]byte, error) {
+	fixedresourceBytes, err := k8s.EncodeResource(fixedResource)
```
klautcomputing: `fixedResourceBytes`?

fix.go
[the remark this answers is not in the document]
genevieveluyt: I think you're right that it is more performant. I was thinking that a manifest probably wouldn't have that many resources, say 10 on average (not based on any stats, just a guess) so looping once vs 3 times wouldn't make much of a difference. I think I prefer the "easier to read" version in this case, but let's get a third opinion. @catherinejones what do you think?

internal/k8s/client.go
klautcomputing: I think this shouldn't be done in this PR, this PR is already overwhelming as it is :) let's create an issue and come back to it in a different PR

internal/k8s/client.go
klautcomputing: Again, open an issue do it in a different PR?

genevieveluyt:
@klautcomputing this is a branch off of `release` which I think is what you are referring to as `kubeaudit-next`? I think it might be nice to squash commit this (and the other auditor branches) and then PR `release` into master to have an overview of all the changes (and commit that without squashing). Or do you think we should keep the command-line only version of kubeaudit alive?

fix.go
genevieveluyt: It's the fixed resource, but in byte form. What do you think it should be called?

kubeaudit.go
```diff
@@ -0,0 +1,177 @@
...
+// AuditManifest audits the Kubernetes resources in the provided manifest
+func (a *Kubeaudit) AuditManifest(manifest io.Reader) (*Report, error) {
+	manifestBytes, err := ioutil.ReadAll(manifest)
+	if err != nil {
+		return nil, err
+	}
+
+	resources, err := getResourcesFromManifest(manifestBytes)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get resources from manifest")
```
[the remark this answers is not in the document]
genevieveluyt: I just realized we're still building for go 1.12 (in addition to 1.13). Do you think it's worth dropping support for 1.12 just for this?

klautcomputing:
> @klautcomputing this is a branch off of release which I think is what you are referring to as kubeaudit-next? I think it might be nice to squash commit this (and the other auditor branches) and then PR release into master to have an overview of all the changes (and commit that without squashing). Or do you think we should keep the command-line only version of kubeaudit alive?

I think we should make a last release of the code that is currently in master. Something like 0.7.1 and then do what you were suggesting. Finish this work in `release` replace `master` with it once it's done and then release a new kubeaudit version. I don't think the old version has a lot of merit and people can just download the source and the bin from releases. So full speed ahead!

kubeaudit.go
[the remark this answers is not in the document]
catherinejones: That's a good point!

internal/k8s/client.go
catherinejones: Sorry, I thought I responded! I didn't realize they were different structures, ignore me! I'm not even certain this warrants an issue, I thought it was a quick optimization but looking at those structures it does not look it.

internal/k8s/client.go
catherinejones: Sorry, I thought I responded! I didn't realize they were different structures, ignore me! I'm not even certain this warrants an issue, I thought it was a quick optimization but looking at those structures it does not look it.

internal/k8s/client.go
catherinejones: Sorry, I thought I responded! I didn't realize they were different structures, ignore me! I'm not even certain this warrants an issue, I thought it was a quick optimization but looking at those structures it does not look it.

Edit: attached this comment to the wrong thread...

fix.go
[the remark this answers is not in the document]
catherinejones: I think unless we've determined this block of code is causing serious performance issues we should optimize for human processing speed over computer processing speed! We can always merge the two loops later if it's a problem!

internal/k8s/client.go
genevieveluyt: https://github.com/Shopify/kubeaudit/issues/242

PR #252, opened 2020-01-22
Add privileged auditor

genevieveluyt:


auditors/privileged/privileged.go
```diff
@@ -0,0 +1,85 @@
...
+			auditResults = append(auditResults, auditResult)
+		}
+	}
+
+	return auditResults, nil
+}
+
+func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource) *kubeaudit.AuditResult {
+	if isPrivilegedNil(container) {
+		// TODO:  should "warning" results include a fix?
```
klautcomputing: Since the change her doesn't change functionality I think we should just include the fix.

PR #258, opened 2020-02-25
Add CLI commands

genevieveluyt:
Adds support back to for using kubeaudit as a CLI.

I changed a few of the command names but kept the old names as aliases so they will still work. Let me know if you think the new names are bad and I'll consider changing them back :)

Also adds support for using a config file to customize the behaviour of the "Audit all" command. It works for both the CLI command and the package as well. Note that the format is different from the original as I think it's clearer this way. Again please let me know if you think otherwise.

New format:
```
enabledAuditors:
    apparmor: true
    asat: true
    ...
auditors:
    capabilities:
        drop: ["NET_ADMIN", "SETPCAP"]
    image:
        image: "myimage:mytag"
    limits:
        cpu: "750m"
        memory: "500m"
```

Old format:
```
apiVersion: v1
kind: kubeauditConfig
audit: true
spec:
  manifest:
  - path: config/kubernetes/default/*/*.yaml
  capabilities:
    NET_ADMIN: drop
    SETPCAP: drop
    ...
  overrides:
    privilege-escalation: deny
    privileged: deny
    ...
```

Using an array of capabilities to drop allows for any capabilities to be dropped (even ones we don't explicitly recommend)

Using auditor names instead of override labels makes it easier to understand what behaviour is or isn't blocked since these names match those in the CLI  docs (`kubeaudit --help`) vs the override labels which were documented in the README.


fix_test.go
```diff
@@ -23,11 +24,14 @@ func TestFix(t *testing.T) {
 		{"all-resources_v1beta2.yml", "all-resources-fixed_v1beta2.yml"},
 	}
 
+	allAuditors, err := all.Auditors(config.KubeauditConfig{})
+	assert.NoError(t, err)
```
catherinejones: If you're depending on this further down in the tests, it might make more sense to use `require.NoError` instead of `assert.NoError`

cmd/commands/rootfs.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/Shopify/kubeaudit/auditors/rootfs"
+	"github.com/spf13/cobra"
+)
+
+var readonlyfsCmd = &cobra.Command{
+	Use:   "rootfs",
+	Short: "Audit containers not using a read only root filesystems",
+	Long: `This command determines which containers have
```
catherinejones: Which containers have what?

genevieveluyt: Oops :)

genevieveluyt:
> Regarding the configuration changes, will this necessitate us updating configuration files in other projects, to support the GitHub check?

Yes it will. The Github check is only enabled on 7 repos, and of those 4 of them use a config file, and of those 4 we own 3 of the repos so it shouldn't be too difficult. Good point to keep in mind though! Here's a list for future reference:

https://github.com/Shopify/online-store-web/blob/master/.github/kubeaudit.yml
https://github.com/Shopify/oscar-the-grouch/blob/master/.github/kubeaudit.yml
https://github.com/Shopify/kubeaudit-webapp/blob/master/.github/kubeaudit.yml
https://github.com/Shopify/security-auditors/blob/master/.github/kubeaudit.yml

PR #261, opened 2020-03-09
Add docs!

genevieveluyt:
##### Description

This PR adds documentation for the package as well as all the auditors and CLI commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 # kubeaudit :cloud: :lock: :muscle:
 
 `kubeaudit` is a command line tool and a Go package to audit Kubernetes clusters for various
-different security concerns:
+different security concerns, such as:
 * run as non-root
 * use a read-only root filesystem
 * drop scary capabilities, don't add new ones
@@ -35,7 +35,7 @@ The rest of this README will focus on how to use kubeaudit as a command line too
  
 ### Download a binary
 
-Kubeaudit has official releases that are blessed and stable here:
+Kubeaudit has official releases that are blessed and stable:
 [Official releases](https://github.com/Shopify/kubeaudit/releases)
 
 ### DIY build
@@ -81,13 +81,14 @@ kubeaudit has three modes:
 
 ### Manifest Mode
 
-If a Kubernetes manifest file is provided using the `-f/--manifest` flag, kubeaudit will audit the manifest file:
+If a Kubernetes manifest file is provided using the `-f/--manifest` flag, kubeaudit will audit the manifest file.
 
+Example command:
 ```
 kubeaudit all -f "/path/to/manifest.yml"
 ```
 
-Example with output:
+Example output:
 ```
 $ kubeaudit all -f auditors/all/fixtures/audit_all_v1.yml
 ERRO[0000] AppArmor annotation missing. The annotation 'container.apparmor.security.beta.kubernetes.io/fakeContainerSC' should be added.  AuditResultName=AppArmorAnnotationMissing Container=fakeContainerSC MissingAnnotation=container.apparmor.security.beta.kubernetes.io/fakeContainerSC
@@ -123,15 +124,14 @@ kubeaudit autofix -f "/path/to/manifest.yml" -o "/path/to/fixed"
 If a kubeconfig file is provided using the `-c/--kubeconfig` flag, kubeaudit will audit the resources specified in the kubeconfig file. If no kubeconfig file is specified, `$HOME/.kube/config` is used by default:
 
 ```
-kubeaudit all --kubeconfig "/path/to/config"
+kubeaudit all -c "/path/to/config"
 ```
 
 For more information on kubernetes config files, see https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/
 
 ### Cluster Mode
 
-If kubeaudit detects it is running within a container, it will try to audit the cluster it is contained in:
-
+Kubeaudit can detect if it is running within a container in a cluster. If so, it will try to audit all Kubernetes resources in that cluster:
 ```
 kubeaudit all
 ```
--- a/README.md
+++ b/README.md
@@ -2,44 +2,54 @@
 [![codecov](https://codecov.io/gh/Shopify/kubeaudit/branch/master/graph/badge.svg)](https://codecov.io/gh/Shopify/kubeaudit)
 [![Go Report Card](https://goreportcard.com/badge/github.com/Shopify/kubeaudit)](https://goreportcard.com/report/github.com/Shopify/kubeaudit)
 [![GoDoc](https://godoc.org/github.com/Shopify/kubeaudit?status.png)](https://godoc.org/github.com/Shopify/kubeaudit)
+
+> Kubeaudit can now be used as both a command line tool (CLI) and as a Go package!
+
 # kubeaudit :cloud: :lock: :muscle:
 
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -1,717 +1,206 @@
-[![Build Status](https://api.travis-ci.org/Shopify/kubeaudit.svg?branch=master)](https://travis-ci.org/Shopify/kubeaudit/)
```
klautcomputing: Can these lines please stay in?

README.md
```diff
@@ -1,717 +1,206 @@
...
+| `capabilities`   | Finds containers that do not drop the recommended capabilities.          | [docs](docs/auditors/capabilities.md) |
+| `hostns`         | Finds containers that have HostPID, HostIPC or HostNetwork enabled.      | [docs](docs/auditors/hostns.md) |
+| `image`          | Finds containers which do not use the desired version of an image (via the tag). | [docs](docs/auditors/image.md) |
+| `limits`         | Finds containers which exceed the specified CPU and memory limits.       | [docs](docs/auditors/limits.md) |
+| `mountds`        | Finds containers that have docker socket mounted.                        | [docs](docs/auditors/mountds.md) |
+| `netpols`        | Finds namespaces that do not have a default-deny network policy.         | [docs](docs/auditors/netpols.md) |
+| `nonroot`        | Finds containers running as root.                                        | [docs](docs/auditors/nonroot.md) |
+| `privesc`        | Finds containers that allow privilege escalation.                        | [docs](docs/auditors/privesc.md) |
+| `privileged`     | Finds containers that have docker socket mounted.                        | [docs](docs/auditors/privileged.md) |
```
klautcomputing: I am not sure the text here is right, also I think we are missing the one for the docker socket :)

README.md
```diff
@@ -1,717 +1,206 @@
...
+| `rootfs`         | Finds namespaces that do not have a default-deny network policy.         | [docs](docs/auditors/rootfs.md) |
```
klautcomputing: yeah something

README.md
```diff
@@ -1,717 +1,206 @@
...
+| `seccomp`        | Finds containers running as root.                                        | [docs](docs/auditors/seccomp.md) |
```
klautcomputing: it wrong here with the descriptions

README.md
genevieveluyt: Oops!

kubeaudit.go
```diff
@@ -1,3 +1,100 @@
...
+//     ...
+//   }
+//
+//   report, err := kubeAuditor.AuditManifest(manifest)
+//
+// Or, to run the audit in local mode:
+//
+//   report, err := kubeAuditor.AuditLocal("/path/to/kubeconfig.yml")
+//
+// Or, to run the audit in cluster mode (pass it a namespace to only audit resources in that namespace):
```
catherinejones:
```suggestion
// Or, to run the audit in cluster mode (pass it a namespace name as a string to only audit resources in that namespace, or an empty string to audit resources in all namespaces):
```

docs/auditors/image.md
```diff
@@ -0,0 +1,41 @@
+# Image Auditor (image)
```
klautcomputing: Should we mention that this isn't on by default when you run `all`?

docs/auditors/limits.md
```diff
@@ -0,0 +1,48 @@
+# Limits Auditor (limits)
+
+Finds containers which exceed the specified CPU and memory limits.
```
klautcomputing:
```suggestion
Finds containers which exceed the specified CPU and memory limits or do not specify any.
```

docs/auditors/privesc.md
```diff
@@ -0,0 +1,69 @@
...
+```
+$ kubeaudit privesc -f "auditors/privesc/fixtures/allow_privilege_escalation_nil_v1.yml"
+ERRO[0000] allowPrivilegeEscalation not set which allows privilege escalation. It should be set to 'false'.  AuditResultName=AllowPrivilegeEscalationNil Container=fakeContainerAPE
+```
+
+## Explanation
+
+`allowPrivilegeEscalation` controls whether a process can gain more privileges than its parent process.
+
+Privilege escalation is disabled by setting `allowPrivilegeEscalation` to `false` in the container SecurityContext. The field defaults to `true` so it must be explicitly set to `false` to pass the `privesc` audit:
```
klautcomputing:
```
By default, ie when allowPrivilegeEscalation=nil, we will set no_new_privs=true with the following exceptions:

when a container is privileged
when CAP_SYS_ADMIN is added to a container
when a container is not run as root, uid 0 (to prevent breaking suid binaries)
```
this is complicated 😄

docs/autofix.md
```diff
@@ -0,0 +1,30 @@
+# Autofix (autofix)
```
klautcomputing: I think this is one of the most awesome features we have! I think we want to hype this a little more 😄

docs/auditors/image.md
genevieveluyt: It does get run by `all` but just only checks for images missing tags

docs/auditors/privesc.md
genevieveluyt: Was that a copy paste from somewhere? What is `no_new_privs`?

docs/autofix.md
genevieveluyt: I agree! Do you have any suggestions?

docs/auditors/privesc.md
klautcomputing: Oh I just have went down a rabbit hole with the documentation you linked 🙂

klautcomputing: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/no-new-privs.md

docs/autofix.md
klautcomputing: How about showing how it keeps comments and how it fixes a basically empty file into a secure one

docs/auditors/privesc.md
genevieveluyt: Ah I see. Do you think we should change the documentation to capture those details?

genevieveluyt: Ah I see. Do you think we should include more detail in our documentation?

klautcomputing: Nah I was just interested in how it works :)

docs/all.md
```diff
@@ -0,0 +1,131 @@
...
+| :------ | :--------- | :---------------------------------------- | :--------------------------------------- |
+| -k      | --kconfig  | Path to kubeaudit config                  |                                          |
+
+Also see [Global Flags](/README.md#global-flags)
+
+## General Usage
+
+```
+kubeaudit all [flags]
+```
```
gurpreetg: I think we should swap the order of `Flags` and `General Usage` section

docs/all.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Kubeaudit Config
+
+The kubeaudit config can be used to specify which auditors to enable and to specify configuration for those auditors. Any configuration that can be specified using flags for the individual auditors can be represented using the config.
```
gurpreetg: This sentence is kind of confusing. We can do something like

```
The kubeaudit config can be used for two things:
1. Enabling specified auditors.
2. Specifying configuration files for the enabled auditors.

Any configuration that can be specified using flags for the individual auditors can be represented using the config.
```

PR #262, opened 2020-03-13
Fix bug where flags were used before initialization

genevieveluyt:
Fixes a bug where flag values for subcommands were being used before they were initialized so passing flags had no effect. Also renaming the `verbose` flag to `minSeverity`

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -17,11 +17,11 @@ import (
 var rootConfig rootFlags
 
 type rootFlags struct {
-	json       bool
-	kubeConfig string
-	manifest   string
-	namespace  string
-	verbose    string
+	json        bool
+	kubeConfig  string
+	manifest    string
+	namespace   string
+	minSeverity string
 }
 
 // RootCmd defines the shell command usage for kubeaudit.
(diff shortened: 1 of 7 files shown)


cmd/commands/util.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Implements pflag.Value
+// Custom type so slice values can be represented as a space-separated list of strings in the flag value
+//
+// Example:
+//
+//     var vals []string
+//     defaultVals := []string{"default1", "default2"}
+//
+//     Flags().VarP(newStringSliceValue(strings.Join(defaultVals, " "), &vals), "flagname", "f", "Flag description")
```
catherinejones: Why space separated and not comma separated?

cmd/commands/root.go
```diff
@@ -31,10 +31,10 @@ var RootCmd = &cobra.Command{
 	Long: `kubeaudit is a program that makes sure all your containers are secure #patcheswelcome
 
 kubeaudit has three modes:
-1. Manifest mode: If a Kubernetes manifest file is provided using the --manifest/-f flag, kubeaudit will audit the manifest file.
+1. Manifest mode: If a Kubernetes manifest file is provided using the -f/--manifest flag, kubeaudit will audit the manifest file.
```
catherinejones: I might throw in a note about it updating the manifest unless you specify an output value here!

cmd/commands/util.go
genevieveluyt: I arbitrarily picked space-separated. Is comma separated the common thing to do?

catherinejones: I'd guess that it's more common! I've never really seen space separated flags tbh!

genevieveluyt: Ok I'll switch it to comma-separated!

cmd/commands/root.go
genevieveluyt: Actually this is just talking about the audit, not the autofix, so it doesn't modify the file. Do you think I should add a note about autofix?

catherinejones: Yeah, I think that makes more sense!

catherinejones: You could always throw that in a follow up PR, I don't want to block this one!

PR #286, opened 2020-08-12
Produce warning if apparmor annotation is invalid

genevieveluyt:
##### Description

If an apparmor annotation references a container that doesn't exist, the manifest will fail to apply. This PR adds a warning for this case.

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [ ] Test A
- [ ] Test B

##### Checklist:

### DIFF
--- a/auditors/apparmor/apparmor.go
+++ b/auditors/apparmor/apparmor.go
@@ -109,7 +109,7 @@ func auditPodAnnotations(resource k8stypes.Resource, containerNames []string) []
 		if !contains(containerNames, containerName) {
 			auditResults = append(auditResults, &kubeaudit.AuditResult{
 				Name:     AppArmorInvalidAnnotation,
-				Severity: kubeaudit.Warn,
+				Severity: kubeaudit.Error,
 				Message:  fmt.Sprintf("AppArmor annotation key refers to a container that doesn't exist. Remove the annotation '%s: %s'.", annotationKey, annotationValue),
 				Metadata: kubeaudit.Metadata{
 					"Container":  containerName,
--- a/auditors/apparmor/apparmor.go
+++ b/auditors/apparmor/apparmor.go
@@ -17,6 +17,9 @@ const (
 	AppArmorAnnotationMissing = "AppArmorAnnotationMissing"
 	// AppArmorDisabled occurs when the apparmor annotation is set to a bad value
 	AppArmorDisabled = "AppArmorDisabled"
+	// AppArmorInvalidAnnotation occurs when the apparmor annotation key refers to a container which doesn't exist. This will
+	// prevent the manifest from being applied to a cluster with AppArmor enabled.
+	AppArmorInvalidAnnotation = "AppArmorInvalidAnnotation"
 )
 
 // As of Jan 14, 2020 these constants are not in the K8s API package, but once they are they should be replaced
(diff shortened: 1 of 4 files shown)


auditors/apparmor/apparmor.go
```diff
@@ -41,14 +44,18 @@ func New() *AppArmor {
...
 	for _, container := range k8s.GetContainers(resource) {
+		containerNames = append(containerNames, container.Name)
 		auditResult := auditContainer(container, resource)
 		if auditResult != nil {
 			auditResults = append(auditResults, auditResult)
 		}
 	}
 
+	auditResults = append(auditResults, auditPodAnnotations(resource, containerNames)...)
+
```
dani-santos-code, line 58: nice unpacking!

auditors/apparmor/apparmor.go
```diff
@@ -92,6 +99,31 @@ func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource)
 	return nil
 }
 
+func auditPodAnnotations(resource k8stypes.Resource, containerNames []string) []*kubeaudit.AuditResult {
+	var auditResults []*kubeaudit.AuditResult
+	for annotationKey, annotationValue := range k8s.GetAnnotations(resource) {
+		if !strings.HasPrefix(annotationKey, ContainerAnnotationKeyPrefix) {
```
dani-santos-code, line 105: oh, I din't know `HasPrefix`. Very useful!

auditors/apparmor/apparmor.go
```diff
@@ -92,6 +99,31 @@ func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource)
...
+			continue
+		}
+		containerName := strings.Split(annotationKey, "/")[1]
+		if !contains(containerNames, containerName) {
+			auditResults = append(auditResults, &kubeaudit.AuditResult{
+				Name:     AppArmorInvalidAnnotation,
+				Severity: kubeaudit.Warn,
```
dani-santos-code, line 112: Could we have `kubeaudit.Error`? kubeuadit will remove that annonation right (`&fix.ByRemovingPodAnnotation`)?

auditors/apparmor/apparmor.go
```diff
@@ -110,3 +142,12 @@ func getProfileName(apparmorAnnotation string, annotations map[string]string) st
...
 }
+
+func contains(arr []string, val string) bool {
+	for _, arrVal := range arr {
+		if arrVal == val {
+			return true
+		}
+	}
+	return false
+}
```
dani-santos-code, line 153: nice helper function!

auditors/apparmor/apparmor.go
genevieveluyt: Good catch! We've struggled with the severity level in the past because we don't have a definition for what should go in each category. Based on the current behaviour I think it could be described as

Error: A security issue
Warn: Best practice and also when Error results are overridden (maybe we should make a separate severity level for overridden results???)
Info: Currently only used by the image auditor to give "success" output (arguably we should get rid of this to be consistent with the rest of kubeaudit where no output means everything is good)

I would say this check isn't directly a security issue nor a best practice 🤔. But I think you're right in that it makes sense to be an error level result rather than a warning!

As a side note, we do have another auditor which produces a warning result with a pending fix. The severity level doesn't matter for autofixing.

dani-santos-code: Yes, you're right. This check doesn't seem to fit either category (Error / Warn) 

What if we use `Info` for overriden results? Or `Notice`? 

Oh good to know! So autofixing doesn't relate to the severity level. Thanks for explaining! 🙏🏽

docs/auditors/apparmor.md
```diff
@@ -17,6 +17,13 @@ $ kubeaudit apparmor -f "auditors/apparmor/fixtures/apparmor_annotation_missing_
 ERRO[0000] AppArmor annotation missing. The annotation 'container.apparmor.security.beta.kubernetes.io/AAcontainer' should be added.  AuditResultName=AppArmorAnnotationMissing Container=AAcontainer MissingAnnotation=container.apparmor.security.beta.kubernetes.io/AAcontainer
 ```
 
+If an apparmor annotation refers to a container which doesn't exist, `kubectl apply` will fail. Kubeaudit produces an error for this case:
+
+```
+$ kubeaudit apparmor -f "auditors/apparmor/fixtures/apparmor-invalid-annotation.yml"
+ERRO[0000] AppArmor annotation key refers to a container that doesn't exist. Remove the annotation 'container.apparmor.security.beta.kubernetes.io/container2: runtime/default'.  Annotation="container.apparmor.security.beta.kubernetes.io/container2: runtime/default" AuditResultName=AppArmorInvalidAnnotation Container=container
+```
```
dani-santos-code, line 25: nice! 👏

PR #287, opened 2020-08-12
Pretty Print

genevieveluyt:
##### Description

image

⚠️  Breaking Change:
Changed the Report.PrintResults function to take options instead of parameters. The default is to print to stdout with minSeverity set to Info (the lowest severity). You can still set a formatter to use logrus (eg. for JSON output). Coloured output is only used if the writer is stdout on a Mac or Linux. Windows does not support coloured terminal output.

⚠️  Breaking Change:
Replaced the `--json` flag with `--format` which can be set to `pretty`, `logrus` (the old default), or `json`

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [ ] Test A
- [ ] Test B

##### Checklist:

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -55,7 +55,7 @@ func init() {
 }
 
 // KubeauditLogLevels represents an enum for the supported log levels.
-var KubeauditLogLevels = map[string]int{
+var KubeauditLogLevels = map[string]kubeaudit.SeverityLevel{
 	"ERROR": kubeaudit.Error,
 	"WARN":  kubeaudit.Warn,
 	"INFO":  kubeaudit.Info,
@@ -65,14 +65,15 @@ func runAudit(auditable ...kubeaudit.Auditable) func(cmd *cobra.Command, args []
 	return func(cmd *cobra.Command, args []string) {
 		report := getReport(auditable...)
 
-		minSeverity := KubeauditLogLevels[rootConfig.minSeverity]
-
-		var formatter log.Formatter
-		if rootConfig.json {
-			formatter = &log.JSONFormatter{}
+		printOptions := []kubeaudit.PrintOption{
+			kubeaudit.WithMinSeverity(KubeauditLogLevels[rootConfig.minSeverity]),
 		}
 
-		report.PrintResults(os.Stdout, minSeverity, formatter)
+		if rootConfig.json {
+			printOptions = append(printOptions, kubeaudit.WithFormatter(&log.JSONFormatter{}))
+		}
+
+		report.PrintResults(printOptions...)
 
 		if report.HasErrors() {
 			os.Exit(2)
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"os"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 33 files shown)


kubeaudit.go
```diff
@@ -216,6 +215,26 @@ func (r *Report) Results() []Result {
...
+		for _, auditResult := range result.GetAuditResults() {
+			if auditResult.Severity >= minSeverity {
+				filteredAuditResults = append(filteredAuditResults, auditResult)
+			}
+		}
+		if len(filteredAuditResults) > 0 {
+			results = append(results, &workloadResult{
+				Resource:     result.GetResource(),
+				AuditResults: filteredAuditResults,
+			})
```
genevieveluyt, line 232: Not sure about this part 🤔

internal/color/color.go
```diff
@@ -0,0 +1,63 @@
...
+		Reset = ""
+		RedColor = ""
+		GreenColor = ""
+		YellowColor = ""
+		BlueColor = ""
+		PurpleColor = ""
+		CyanColor = ""
+		GrayColor = ""
+		WhiteColor = ""
+	}
```
dani-santos-code, line 62: What does this means? no colors for windows? 😅

kubeaudit.go
```diff
@@ -216,6 +215,26 @@ func (r *Report) Results() []Result {
 	return results
 }
 
+// Results returns the audit results for each Kubernetes resource with a minimum severity
```
dani-santos-code, line 218: `// ResultsWithMinSeverity returns the audit results for each Kubernetes resource with a minimum severity`

kubeaudit.go
dani-santos-code: It looks good to me.  What is it that makes you unsure? 🤔

printer.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (p *Printer) prettyPrintReport(report *Report) {
+	for _, workloadResult := range report.ResultsWithMinSeverity(p.minSeverity) {
+		resource := workloadResult.GetResource().Object()
+		groupVersionKind := resource.GetObjectKind().GroupVersionKind()
+		resourceName := k8s.GetObjectMeta(resource).GetName()
+		resourceNamespace := k8s.GetObjectMeta(resource).GetNamespace()
+
+		p.printColor(color.CyanColor, "\n--------- Results for ---------------------\n\n")
```
dani-santos-code, line 74: nitpick! can we me make it more centralized? 😬

`"\n---------------- Results for ---------------\n\n"
`
(I counted so it gets centralized like this. 17 to the left, 17 to the right)

result.go
```diff
@@ -5,13 +5,13 @@ import "github.com/Shopify/kubeaudit/k8stypes"
 // AuditResult severity levels. They also correspond to log levels
 const (
 	// Info is used for informational audit results where no action is required
-	Info = iota
+	Info SeverityLevel = 0
```
dani-santos-code, line 8: nice! 👏 it's more readable!

result.go
```diff
@@ -20,13 +20,28 @@ type Result interface {
 	GetAuditResults() []*AuditResult
 }
 
+type SeverityLevel int
```
dani-santos-code, line 23: 👏

cmd/commands/root.go
```diff
@@ -49,13 +49,13 @@ func Execute() {
 func init() {
 	RootCmd.PersistentFlags().StringVarP(&rootConfig.kubeConfig, "kubeconfig", "c", "", "Path to local Kubernetes config file. Only used in local mode (default is $HOME/.kube/config)")
 	RootCmd.PersistentFlags().StringVarP(&rootConfig.minSeverity, "minseverity", "m", "INFO", "Set the lowest severity level to report (one of \"ERROR\", \"WARN\", \"INFO\")")
-	RootCmd.PersistentFlags().BoolVarP(&rootConfig.json, "json", "j", false, "Output audit results in JSON")
+	RootCmd.PersistentFlags().StringVar(&rootConfig.format, "format", "pretty", "The output format to use (one of \"pretty\", \"logrus\", \"json\")")
 	RootCmd.PersistentFlags().StringVarP(&rootConfig.namespace, "namespace", "n", apiv1.NamespaceAll, "Only audit resources in the specified namespace. Not currently supported in manifest mode.")
```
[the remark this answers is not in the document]
genevieveluyt: Sure

cmd/commands/root.go
```diff
@@ -65,14 +65,18 @@ func runAudit(auditable ...kubeaudit.Auditable) func(cmd *cobra.Command, args []
 	return func(cmd *cobra.Command, args []string) {
 		report := getReport(auditable...)
 
-		minSeverity := KubeauditLogLevels[rootConfig.minSeverity]
+		printOptions := []kubeaudit.PrintOption{
+			kubeaudit.WithMinSeverity(KubeauditLogLevels[rootConfig.minSeverity]),
+		}
```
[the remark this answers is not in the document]
genevieveluyt: Yea it's always been a capitalized flag value as is specified in all the docs, but we could change it to be lowercase to be consistent with the rest of the flags

internal/color/color.go
genevieveluyt: Windows doesn't support terminal colours unfortunately

kubeaudit.go
genevieveluyt: I think it would be ideal to copy the original object and modify only the audit results rather than creating a new object from scratch, because now we have to specify which fields to copy (in this case only `Resource` but maybe we add more fields in the future). We can't copy the original object though because it's an interface.

PR #293, opened 2020-09-01
Fix function naming

genevieveluyt:
##### Description

Fixes function naming. This is basically a no-op.


genevieveluyt:
Closing in favour of #296

PR #297, opened 2020-09-03
Support automountServiceAccountToken on ServiceAccount

### LINKED ISSUE #290: K8s 1.6+: Incorrect 'Default service account with token mounted'
##### ISSUE TYPE
 - [x] Bug Report

# BUG REPORT

##### SUMMARY

kubeaudit reports `AutomountServiceAccountTokenTrueAndDefaultSA` even if the service account spec has `"automountServiceAccountToken": false`, as available in K8s 1.6+ (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server)

##### ENVIRONMENT
* Kubeaudit version: v0.9.0
* Kubeaudit install method: DIY build

##### STEPS TO REPRODUCE

- Notice that by default, the `default` service account token is mounted in pods

```
$ kubectl run --rm -i --tty shell --image=alpine --restart=Never -- ls /var/run/secrets/kubernetes.io/serviceaccount/
ca.crt     namespace  token
```

- Patch the default service account so it doesn't automatically mount service account tokens in pods:

```
$ kubectl patch serviceaccount default -p '{"automountServiceAccountToken": false}'
```

- Start a pod, and notice the service account token isn't mounted anymore:

```
$ kubectl run --rm -i --tty shell --image=alpine --restart=Never -- ls /var/run/secrets/kubernetes.io/serviceaccount/
ls: /var/run/secrets/kubernetes.io/serviceaccount/: No such file or directory
```

- Notice that `kubeaudit asat` still complains about a service account token being automatically mounted:

```
$ kubeaudit asat
ERRO[0000] Default service account with token mounted. automountServiceAccountToken should be set to 'false' or a non-default service account should be used.  AuditResultName=AutomountServiceAccountTokenTrueAndDefaultSA
```

##### EXPECTED RESULTS

No output from kubeaudit, as the default service account token is not mounted

genevieveluyt:
##### Description

Makes it so the `asat` auditor won't complain that token automounting is enabled if `automountServiceAccountToken: false` is set on the default ServiceAccount (and `automountServiceAccountToken` is not set to `true` in the PodSpec).

Fixes #290 

##### Type of change

- [x] Bug fix :bug:
- [ ] New feature :sparkles:

- [x] If the default service account has `automountServiceAccountToken: false` and the PodSpec does not have `automountServiceAccountToken: true` => no error
- [x] In manifest mode, if the default service account is being automounted and we know about the default service account, autofix adds `automountServiceAccountToken: true` to the default service account, otherwise on the PodSpec

##### Checklist:

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -304,7 +304,7 @@ func GetNetworkPolicies(clientset kubernetes.Interface, options ClientOptions) [
 	return netPols
 }
 
-// GetCronJobs gets all CronJob resources from the cluster
+// GetServiceAccounts gets all ServiceAccount resources from the cluster
 func GetServiceAccounts(clientset kubernetes.Interface, options ClientOptions) []k8stypes.Resource {
 	serviceAccountClient := clientset.CoreV1().ServiceAccounts(options.Namespace)
 	serviceAccountList, err := serviceAccountClient.List(context.Background(), k8stypes.ListOptionsV1{})
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -96,6 +96,7 @@ func GetAllResources(clientset kubernetes.Interface, options ClientOptions) []k8
 	resources = append(resources, GetStatefulSets(clientset, options)...)
 	resources = append(resources, GetNetworkPolicies(clientset, options)...)
 	resources = append(resources, GetCronJobs(clientset, options)...)
+	resources = append(resources, GetServiceAccounts(clientset, options)...)
 	resources = append(resources, GetNamespaces(clientset, options)...)
 
 	resources = excludeGenerated(resources)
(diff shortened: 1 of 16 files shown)


internal/k8s/client.go
```diff
@@ -303,6 +304,29 @@ func GetNetworkPolicies(clientset kubernetes.Interface, options ClientOptions) [
 	return netPols
 }
 
+// GetCronJobs gets all CronJob resources from the cluster
```
maaslalani, line 307:
```suggestion
// GetServiceAccounts gets all ServiceAccounts resources from the cluster
```

k8stypes/resources.go
```diff
@@ -125,3 +125,14 @@ func NewCronJob() *CronJobV1Beta1 {
...
 }
+
+// NewServiceAccount creates a new ServiceAccount resource
+func NewServiceAccount() *ServiceAccountV1 {
+	serviceAccount := &ServiceAccountV1{
+		ObjectMeta: ObjectMetaV1{},
+	}
+
+	serviceAccount.Kind = "ServiceAccount"
+	serviceAccount.APIVersion = "v1"
```
maaslalani, line 136: Is there a difference between setting the `Kind` and `APIVersion` this way vs using the `SetGroupVersionKind` like the `GetServiceAccounts` function?

```
serviceAccount.SetGroupVersionKind(schema.GroupVersionKind{
	Kind:    "ServiceAccount",
	Version: "v1",
})
```

maaslalani (dismissed):
Code, docs and tests look good to me! I 🎩 'd locally and everything seems to work correctly 😄

k8stypes/resources.go
genevieveluyt: I believe they are effectively the same. `SetGroupVersionKind` is implemented on the interface so it can be called on both the generic `k8stypes.Resource` that we pass around in most of kubeaudit and on the specific resource types. Good catch though, I'll clean up the types so they're more consistent.

maaslalani: Sounds good! 👍

maaslalani (approved):

PR #305, opened 2020-09-28
Adds ALL to drop list and removes added capabilities

### LINKED ISSUE #260: Capabilities auditor should drop "all"
##### ISSUE TYPE
 - [ ] Bug Report
 - [x] Feature Idea

# FEATURE IDEA

**Proposal:** Currently, the capabilities auditor drops all capabilities which are added by default
ie.
```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      containers:
      - name: myContainer
        securityContext:
          capabilities:
            drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETUID
            - SETPCAP
            - SYS_CHROOT
```
If capabilities from this list are needed, people need to remove them from the drop list and add an override label so that kubeaudit won't generate errors.

Instead, we should drop all capabilities and allow people to add specific capabilities with an override label:
```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    metadata:
      labels:
        container.audit.kubernetes.io/myContainer.allow-capability-audit-write: ""
    spec:
      containers:
      - name: myContainer
        securityContext:
          capabilities:
            drop:
            - ALL
            add:
            - AUDIT_WRITE
```

dani-santos-code:
##### Description

Closes #260 

One question about the `--allowAddList` flag for the capabilities command. I added the following description: `"Comma separated list of added capabilities that can be ignored by kubeaudit reports"`

Here is the flow:

For each capability added (either to the `add` list or the `drop` list), `kubeaudit` will report errors, unless the capability is overridden with a label. Only one message will be displayed for `CapabilityShouldDropAll`, rather than displaying the same report for each capability, as the list coulg get very long:

Screen Shot 2020-10-28 at 11 28 10 AM

When running `autofix` kubeaudit will set all capabilities in the drop list to "ALL", and delete all capabilities in the add list:

caps_fix

If we pass in a` config `file  followed by the `add` capabilities flag, the flag will take precedence. For example:

Say we have the following added to our add list in `myconfig.yaml`:
```yaml
auditors:
  capabilities:
    add:
      - DAC_OVERRIDE
      - FSETID
      - KILL
      - MKNOD
```
And we have the following in the manifest file`myfile.yaml`:

```yaml
        securityContext:
            capabilities:
              add:
                - SYS_TIME
                - SYS_MODULE
                - DAC_OVERRIDE
                - FSETID
                - KILL
                - MKNOD
```

After running:

```shell
kubeaudit all -k myconfig.yaml -f myfile.yaml --add "SYS_MODULE,CHOWN"
```
The result will be that the only capabilities added to Add List ignored by `kubeaudit` will be the ones listed in the comma separated list ("SYS_MODULE,CHOWN"):

Screen Shot 2020-11-05 at 4 47 32 PM

##### Caveats

- Test coverage for capabilities decreased by nearly 3 % (It used to be 94% now it's 91.3%)

##### Type of change

- [x] New feature :sparkles:

- [x] Automated tests

##### Checklist:


README.md
```diff
@@ -122,8 +122,8 @@ $ kubeaudit all -f "internal/test/fixtures/all_resources/deployment-apps-v1.yml"
 -- [error] AutomountServiceAccountTokenTrueAndDefaultSA
    Message: Default service account with token mounted. automountServiceAccountToken should be set to 'false' or a non-default service account should be used.
 
--- [error] CapabilityNotDropped
-   Message: Capability not dropped. Ideally, the capability drop list should include the single capability 'ALL' which drops all capabilities.
+-- [error] CapabilityShouldDropAll
+   Message: Capabily not set to ALL. Ideally, you should drop ALL capabilities and add the specific ones you need to the add list.
```
catherinejones, line 126: You have a typo here!
```suggestion
   Message: Capability not set to ALL. Ideally, you should drop ALL capabilities and add the specific ones you need to the add list.
```

auditors/capabilities/capabilities.go
```diff
@@ -73,45 +60,70 @@ func getOverrideLabel(capability string) string {
...
+	if isCapabilityInArray(capability, addList) {
+		return auditResults
 	}
 
-	if isCapabilityNotDropped(container, capability, dropList) {
+	if SecurityContextOrCapabilities(container) {
+		var message string
+
+		if IsCapabilityInAddList(container, capability) {
+			message = fmt.Sprintf("Capability added. It should be removed from the capability add list. If you need this capability, add an override label such as '%s: SomeReason'.", override.GetContainerOverrideLabel(container.Name, getOverrideLabel(capability)))
```
catherinejones, line 74: Would it be possible to specify the capability that is Added? Something like:
```suggestion
			message = fmt.Sprintf("Capability \"%s\" added. It should be removed from the capability add list. If you need this capability, add an override label such as '%s: SomeReason'.", capability, override.GetContainerOverrideLabel(container.Name, getOverrideLabel(capability)))
```

auditors/capabilities/capabilities.go
```diff
@@ -73,45 +60,70 @@ func getOverrideLabel(capability string) string {
...
+				},
+				Metadata: kubeaudit.Metadata{
+					"Container": container.Name,
+				},
+			}
+			auditResults = append(auditResults, auditResult)
+		}
+
+		if !IsDropAll(container) && !IsCapabilityInAddList(container, capability) {
+			message = "Capabily not set to ALL. Ideally, you should drop ALL capabilities and add the specific ones you need to the add list."
```
catherinejones, line 91: This message is a little less clear to me, might I suggest:

```suggestion
			message = "Capability Drop list is not set to ALL. Ideally, you should drop ALL capabilities and add the specific ones you need to the Add list."
```

auditors/capabilities/capabilities.go
```diff
@@ -73,45 +60,70 @@ func getOverrideLabel(capability string) string {
...
+					capability: capability,
+				},
+				Metadata: kubeaudit.Metadata{
+					"Container": container.Name,
+				},
+			}
+			auditResults = append(auditResults, auditResult)
+		}
+	} else {
+		message := "Security Context not set. Ideally, the Security Context should be specified. All capacities should be dropped by setting drop to ALL."
```
catherinejones, line 107:
```suggestion
		message := "Security Context not set. Ideally, the Security Context should be specified and all Capabilities should be dropped by setting the Drop list to ALL."
```

catherinejones (changes requested):
Looks good to me! I have a few wording changes to request!

dani-santos-code:
Thanks for your suggestions and review, @catherinejones! 🙏🏽 I applied your suggestions! 😊

rxbchen (approved):
Looks good, just one small change

auditors/capabilities/fix.go
```diff
@@ -63,3 +62,24 @@ func removeCapabilityFromAddList(container *k8stypes.ContainerV1, capability str
...
+	return fmt.Sprintf("Adds capability to '%s' with drop set to ALL", f.container.Name)
+}
+
+func (f *fixMissingSecurityContextOrCapability) Apply(resource k8stypes.Resource) []k8stypes.Resource {
+	setDropToAll(f.container)
+	return nil
+}
+
+func setDropToAll(container *k8stypes.ContainerV1) {
+	container.SecurityContext = &k8stypes.SecurityContextV1{}
```
genevieveluyt, line 80: If the container already has a security context I believe this will blow it away (we should only modify the capabilities)

docs/auditors/capabilities.md
```diff
@@ -49,47 +57,109 @@ $ kubeaudit capabilities -f "auditors/capabilities/fixtures/capabilities-nil.yml
...
+    name: deployment
+    namespace: capabilities-some-allowed-multi-containers-some-labels
 
+--------------------------------------------
+
+-- [error] CapabilityAdded
+   Message: Capability added. It should be removed from the capability add list. If you need this capability, add an override label such as 'container.audit.kubernetes.io/container1.allow-capability-net-admin: SomeReason'.
+   Metadata:
+      Container: container1
+
```
genevieveluyt, line 160: Can we include the capability in the Metadata?

auditors/capabilities/fix.go
```diff
@@ -63,3 +62,24 @@ func removeCapabilityFromAddList(container *k8stypes.ContainerV1, capability str
 
 	container.SecurityContext.Capabilities.Add = added
 }
+
+type fixMissingSecurityContextOrCapability struct {
+	container *k8stypes.ContainerV1
+}
+
+func (f *fixMissingSecurityContextOrCapability) Plan() string {
+	return fmt.Sprintf("Adds capability to '%s' with drop set to ALL", f.container.Name)
```
genevieveluyt, line 71: I'm a bit confused by the `Adds capability to '%s'`. This sounds like we're adding a capability. Maybe it can just be something like `Sets capability drop list  to ["ALL"]`?

dani-santos-code:
Thanks, @rxbchen ! 🙏

genevieveluyt (commented):
This is looking good! I will continue reviewing after my meeting 🙂

auditors/capabilities/fix.go
dani-santos-code: good catch!! I will change it!

auditors/capabilities/fix.go
dani-santos-code: That was a bad copy pasta from another fix! 🙈 Thanks for catching it! 🙏🏽

auditors/capabilities/fixtures/capabilities-redundant-override.yml
```diff
@@ -11,25 +11,12 @@ spec:
     metadata:
       labels:
         name: deployment
-        audit.kubernetes.io/pod.allow-capability-chown: "SomeReason"
+        audit.kubernetes.io/pod.allow-capability-all: "SomeReason"
```
genevieveluyt, line 14: I don't think we want anyone to ever override the `all` capability. Even though there's nothing in the code stopping someone from doing this, I don't think we should make it an example. Since there's nothing in the add list, any capability override is redundant, so I think keeping it as `audit.kubernetes.io/pod.allow-capability-chown` should be fine

auditors/capabilities/fixtures/capabilities-some-allowed.yml
```diff
@@ -21,18 +21,6 @@ spec:
             capabilities:
               add:
                 - SYS_TIME
-                - SYS_MODULE
+                - CHOWN
```
genevieveluyt, line 24: This was intentionally not matching the override label since it's testing "some allowed"

docs/all.md
```diff
@@ -131,10 +130,13 @@ enabledAuditors:
   # Auditors are enabled by default if they are not explicitly set to "false"
   hostns: false
   image: false
-  limits: false
 auditors:
   capabilities:
-    drop: ['AUDIT_WRITE', 'CHOWN']
+  drop:
```
genevieveluyt, line 135: `drop` is not an option in the config anymore

docs/auditors/capabilities.md
```diff
@@ -49,81 +31,151 @@ $ kubeaudit capabilities -f "auditors/capabilities/fixtures/capabilities-nil.yml
...
+                - CHOWN
+                - KILL
+                - MKNOD
+              drop:
+                - ALL
 ```
-$ kubeaudit capabilities --drop "MAC_ADMIN AUDIT_WRITE" -f "auditors/capabilities/fixtures/capabilities-nil.yml"
+
+```shell
+$ kubeaudit all --kconfig "example.yaml" -f "test.yaml"
```
genevieveluyt, line 86: Can we name the two files more descriptively so this command is easier to understand? eg `config.yaml` and `manifest.yaml`

docs/auditors/capabilities.md
```diff
@@ -49,81 +31,151 @@ $ kubeaudit capabilities -f "auditors/capabilities/fixtures/capabilities-nil.yml
...
+        - name: container1
+          image: scratch
+          securityContext:
+            capabilities:
+              add:
+                - CHOWN
+                - KILL
+                - MKNOD
+              drop:
+                - ALL
```
genevieveluyt, line 82: I think we should make the example yaml files as simple as possible eg.

```suggestion
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  name: deployment
  namespace: capabilities-some-allowed-multi-containers-some-labels
spec:
  template:
    spec:
      containers:
        - name: container1
          image: scratch
          securityContext:
            capabilities:
              add:
                - CHOWN
                - KILL
                - MKNOD
              drop:
                - ALL
```

And maybe use a friendlier namespace name?

auditors/capabilities/capabilities.go
```diff
@@ -12,53 +12,51 @@ import (
...
+		var capabilitiesToDrop []string
+
+		for _, capability := range mergeCapabilities(container) {
+			for _, auditResult := range auditContainer(container, capability, a.addList) {
+				auditResult = override.ApplyOverride(auditResult, container.Name, resource, getOverrideLabel(capability))
+				if auditResult != nil && auditResult.Name == CapabilityShouldDropAll {
+					capabilitiesToDrop = append(capabilitiesToDrop, capability)
+					if len(capabilitiesToDrop) == 1 {
+						auditResults = append(auditResults, auditResult)
+					}
```
genevieveluyt, line 55: The original idea behind calling `auditContainer` per capability is because we wanted to have an audit result per capability per container. But since this is not the case for the `CapabilityShouldDropAll` result (we only want one), we could move that check to a separate function (eg. `auditContainerForDropAll`) and call that outside the loop (rather than doing the work for each capability and discarding all but one). I wonder if we event need to merge capabilities anymore, I think we could just use the add list?

```go
for _, container := range k8s.GetContainers(resource) {
  for _, capability := range GetCapabilityDropList(container) {
    // call auditContainer and add to audit results
  }

  auditResult := auditContainerForDropAll(container)
  // add audit result to results if not nil, don't need to check for override label
}
```

auditors/capabilities/fix.go
```diff
@@ -63,3 +62,32 @@ func removeCapabilityFromAddList(container *k8stypes.ContainerV1, capability str
...
+		container.SecurityContext = &k8stypes.SecurityContextV1{}
+	}
+
+	if container.SecurityContext.Capabilities == nil {
+		container.SecurityContext.Capabilities = &k8stypes.CapabilitiesV1{}
+	}
+
+	if container.SecurityContext.Capabilities.Drop == nil {
+		container.SecurityContext.Capabilities.Drop = []k8stypes.CapabilityV1{}
+	}
```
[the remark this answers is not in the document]
dani-santos-code: true!! 🙏🏽

auditors/capabilities/fix_test.go
```diff
@@ -24,72 +24,80 @@ func TestFixCapabilities(t *testing.T) {
...
+			testName:     "Nothing to fix - no caps added and drop is set to all",
 			overrides:    []string{},
 			add:          []string{},
 			expectedAdd:  []string{},
 			drop:         []string{"all"},
 			expectedDrop: []string{"all"},
 		},
 		{
-			testName:     "CapabilityNotDropped",
+			testName:     "No capabilities specified - nothing to fix",
```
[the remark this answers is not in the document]
dani-santos-code: True!! I will make changes in the test to check for when no caps are specified and all we do is to add all to the drop list!

cjones-shopify (dismissed):
Looks good to me!

auditors/capabilities/capabilities.go
```diff
@@ -73,45 +65,74 @@ func getOverrideLabel(capability string) string {
...
+	if len(auditResults) == 0 {
+		return []*kubeaudit.AuditResult{nil}
+	}
+
+	return auditResults
+}
+
+func auditContainerForDropAll(container *k8stypes.ContainerV1) *kubeaudit.AuditResult {
+	var message string
+
```
genevieveluyt, line 107: This is a bit weird as the message variable on line 109 is a different variable whereas the one on line 124 is using this one. I think we could make them both instantiate (using `:=`) and get rid of this function-scoped variable.

auditors/capabilities/capabilities.go
```diff
@@ -73,45 +65,74 @@ func getOverrideLabel(capability string) string {
...
+	if isCapabilityInArray(capability, addList) {
+		return auditResults
+	}
+
+	if SecurityContextOrCapabilities(container) {
+		var message string
+
+		if IsCapabilityInAddList(container, capability) {
+			message = fmt.Sprintf("Capability \"%s\" added. It should be removed from the capability add list. If you need this capability, add an override label such as '%s: SomeReason'.", capability, override.GetContainerOverrideLabel(container.Name, getOverrideLabel(capability)))
```
genevieveluyt, line 79:
```suggestion
		if IsCapabilityInAddList(container, capability) {
			message := fmt.Sprintf("Capability \"%s\" added. It should be removed from the capability add list. If you need this capability, add an override label such as '%s: SomeReason'.", capability, override.GetContainerOverrideLabel(container.Name, getOverrideLabel(capability)))
```

auditors/capabilities/config.go
```diff
@@ -1,12 +1,13 @@
...
-	DropList []string `yaml:"drop"`
+	AddList []string `yaml:"add"`
 }
 
-func (config *Config) GetDropList() []string {
-	if config == nil || len(config.DropList) == 0 {
-		return DefaultDropList
+func (config *Config) GetAddList() []string {
+	if config == nil || len(config.AddList) == 0 {
+		return []string{}
```
genevieveluyt, line 9: Should this be `DefaultAddList`?

docs/all.md
```diff
@@ -145,26 +145,8 @@ $ kubeaudit all -k "config.yaml" -f "auditors/all/fixtures/audit_all_v1.yml"
 
 ### Example with Flags
 
-The behaviour of the `all` command can also be customized by using flags. The `all` command supports all flags supported by invididual auditors (see the individual [auditor docs](/README.md#auditors) for all the flags). For example, the `caps` auditor supports specifying capabilities to drop with the `--drop/-d` flag so this flag can be used with the `all` command:
+The behaviour of the `all` command can also be customized by using flags. The `all` command supports all flags supported by invididual auditors (see the individual [auditor docs](/README.md#auditors) for all the flags). For example, the `caps` auditor supports specifying capabilities to add with the `--add` flag so this flag can be used with the `all` command
```
genevieveluyt, line 148: I'm not sure `specifying capabilities to add` is a clear description of the flag since it doesn't actually add the capabilities, it just tells kubeaudit to not complain if they are added. It might be worth using a different auditor here, like the `limits` auditor, since the flags are easier to explain.

genevieveluyt (dismissed):
Just some nits but the code looks good! I am having second thoughts about "add list" for the config and flag option. I wonder if we can think of something less confusing. Maybe like "allow list" or even "allowAdd list". Wdyt?

auditors/capabilities/config.go
dani-santos-code: yup! good catch!

auditors/capabilities/capabilities.go
dani-santos-code: Agreed! I probably had something else in mind when I did that, and ended up leaving that function-scoped variable there. I'll remove it!

auditors/capabilities/capabilities.go
dani-santos-code: yeah! same thing here

dani-santos-code:
> Just some nits but the code looks good! I am having second thoughts about "add list" for the config and flag option. I wonder if we can think of something less confusing. Maybe like "allow list" or even "allowAdd list". Wdyt?
@genevieveluyt I like "allowAddlist" and I think it would be less confusing indeed! I'll change it!

docs/all.md
```diff
@@ -145,26 +145,20 @@ $ kubeaudit all -k "config.yaml" -f "auditors/all/fixtures/audit_all_v1.yml"
...
-
-```yaml
-auditors:
-    capabilities:
-        drop: ["CHOWN", "MKNOD]
+kubeaudit all -f "manifest.yml" --memory 200
 ```
 
-The capabilities specified by the flag will take precedence over those specified in the config file resulting in only `AUDIT_WRITE` being dropped.
+Here, if the memory specified is higher than 200, `kubeaudit` will report that the memory limit was exceeded.
```
dani-santos-code, line 164: @genevieveluyt what do you think of rephrasing here and also adding the `limits` as an extra example, as you suggested?

genevieveluyt: I think just one would be sufficient. The more duplicated documentation we have the more likely it'll get out of sync. Wdyt?

genevieveluyt (dismissed):

genevieveluyt (approved):
🚢

cjones-shopify (approved):

Issue #264, opened 2020-03-17
Autofix should use a configuration file


genevieveluyt:
##### ISSUE TYPE
 - [ ] Bug Report
 - [x] Feature Idea

# FEATURE IDEA

**Proposal:**

The `autofix` command should use the configuration set in the kubeaudit config file.

PR #313, opened 2020-10-19
Autofix with config

dani-santos-code:
### Description

Closes https://github.com/Shopify/kubeaudit/issues/264

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [x] Manual tests

##### Checklist:


cmd/commands/autofix.go
```diff
@@ -4,18 +4,30 @@ import (
...
+	kubeauditConfigFile string
 }
 
 func autofix(cmd *cobra.Command, args []string) {
-	report := getReport()
+	conf := loadKubeAuditConfigFromFile(autofixConfig.kubeauditConfigFile)
+
+	conf = setConfigFromFlags(cmd, conf)
+
+	allAuditors, err := all.Auditors(conf)
```
dani-santos-code, line 23: should I just call it `auditors`? or `configEnabledAuditors`?

genevieveluyt: Auditors makes sense to me! This code looks like it's based on the `all` command so I think we should change it there too!

cmd/commands/autofix.go
```diff
@@ -35,20 +47,42 @@ func autofix(cmd *cobra.Command, args []string) {
 	}
 }
 
+func loadKubeAuditConfigFromFile(configFile string) config.KubeauditConfig {
```
genevieveluyt, line 50: This looks like it's based on `loadConfigFromFile` from the `all` command. I think that function was meant to use `configFile` (the passed in parameter) instead of `auditAllConfig.configFile` (since the passed in parameter is not used anywhere in the function 🤦 ). I do like `loadKubeAuditConfigFromFile` better as a function name though. So I think what we should do is keep this function but replace `autofixConfig.kubeauditConfigFile` with `configFile`, then remove the `loadConfigFromFile` function, and have the auditAll command call this function instead. Wdyt?

docs/autofix.md
```diff
@@ -263,3 +264,13 @@ To write the fixed manifest to a different file, use the `--outfile/-o` flag:
 ```
 kubeaudit autofix -f "manifest.yml" -o "fixed.yaml"
 ```
+
+### Using Custom Rules with Kubeaudit Config File
```
genevieveluyt, line 268: Now that two commands support the kubeaudit config maybe we can move the documentation from the `all` command (https://github.com/Shopify/kubeaudit/blob/master/docs/all.md#kubeaudit-config) to the home page (https://github.com/Shopify/kubeaudit#configuration-file) and then link there from the `all` and `autofix` readmes?

genevieveluyt (commented):
Awesome this is very clean!

cmd/commands/autofix.go
dani-santos-code: totally!! it makes sense! I also failed to notice `configFile` wasn't being used! 😂I will change it and also rename the function!

docs/autofix.md
dani-santos-code: good idea!!!

README.md
```diff
@@ -216,7 +222,50 @@ Auditors can also be run individually.
...
+    # will be generated for containers which use an image without a tag
+    image: 'myimage:mytag'
+  limits:
+    # If no limits are specified and the 'limits' auditor is enabled, WARN results
+    # will be generated for containers which have no cpu or memory limits specified
+    cpu: '750m'
+    memory: '500m'
+```
+
+For more details about each auditor, including a description of the auditor-specific configuration in the config, see the [Auditor Docs](/README.md#auditors).
```
genevieveluyt, line 266:
```suggestion
For more details about each auditor, including a description of the auditor-specific configuration in the config, see the [Auditor Docs](#auditors).
```

fix.go
```diff
@@ -1,135 +0,0 @@
-package kubeaudit
```
genevieveluyt, line 1: Why is this file removed?

fixed.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1beta2
```
genevieveluyt, line 1: Was this file added intentionally?

fix.go
dani-santos-code: omg! I didn't mean to delete it! 😩I'll bring it back!

fixed.yaml
dani-santos-code: noooo! sorry! I made a mistake when going back and forth! I should have reviewed this more carefully before asking for feedback!

dani-santos-code: I was supposed to have deleted the yaml and deletes the fix.go`` by accident instead!

genevieveluyt (approved):
LGTM!

cjones-shopify (approved):
This looks good to me!

PR #316, opened 2020-12-30
Add a new command to audit runAsUser fields

jcbbc:
##### Description

Add a new command to audit runAsUser fields. The audit will trigger an alert when the container user ID is not overridden with a non-root user using the `runAsUser` either in the Pod Security Context or the container Security Context. The check will fail if no `runAsUser` is specified or if it uses the `0` UID. This is useful to enforce non-root user in container at runtime.

Here's a sample result:
```
kubeaudit runasuser -f "auditors/runasuser/fixtures/run-as-user-0.yml" 

  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: deployment
    namespace: run-as-user-0

-- [error] RunAsUserCSCRoot
   Message: container user ID not overridden to non-root user using runAsUser SecurityContext. It should be set to > 0.
   Metadata:
      Container: container
```

I've also modified the `nonroot` command's description from `This command determines which containers are running as root (uid=0)` to `This command determines which containers are allowed to run as root (uid=0).` because even when `runAsNonRoot` is set to `true` or is missing, this doesn't mean that the container will effectively run as root, only that it will be permitted to use the root user if it's the one specified in the image. 

##### Type of change
- [x] New feature :sparkles:
- [x] This change requires a documentation update 📖

#####  How Has This Been Tested?
- [x] Automated tests
#####  Checklist:
- [x] I have 🎩 my changes (A 🎩 specifically includes pulling down changes, setting them up, and manually testing the changed features and potential side effects to make sure nothing is broken)
- [] The test coverage did not decrease
- [x] I have signed the appropriate Contributor License Agreement


genevieveluyt (commented):
Hey @jcbbc,

Thanks so much for your thorough PR, the code looks solid and there are an impressive amount of tests!

My main thought is that the `nonroot` and `runasuser` auditors are mutually exclusive, and that we should only run one of them, not both. The security concern is images running as the root user, which is mitigated by either setting `runAsNonRoot: true` or by setting `runAsUser` to a non-root user, and we already check for the first one. If you run "all", you will be required to add both to make all the errors go away.

I can see that one or the other may be more convenient depending on if you set a non-root USER in the majority of your Dockerfiles or not. If you do set USER in most of your Dockerfiles, then setting `runAsNonRoot: true` is better and you would have to override the `runAsUser` check (or disable the auditor). If you don't set a USER, then setting `runAsUser` is better since `runAsNonRoot` would break your deploy. 

I almost think they should be the same auditor to assert "this image is not running as root", which checks that either `runAsNonroot: true` is set or `runAsUser > 0`. We could add a flag to the `autofix` command to specify the user (eg. `--user 1000`) which will then autofix to `runAsUser: 1000` and otherwise it would set `runAsNonRoot: true`.

Alternatively, we could make this more of an informational check (like the `image` auditor), which just produces warnings, or choose one to be the default and have a way to switch to the other, or have it disabled by default.

Mainly, I think we should keep the default behaviour sane (ie. running "all" and "autofix" gives reasonable results for most people) and having errors for both `runAsNonroot` and `runAsUser` feels unnecessary to me.

What do you think?

jcbbc:
> Thanks so much for your thorough PR, the code looks solid and there are an impressive amount of tests!

Thanks! To be honest, I mainly copy/paste and adapt from the `nonroot` auditor which is very well tested.

I agree with you. I didn't think about the `all` auditor but indeed it may not make sense to have one of them them failing if you rely on the other. 

I like your proposal for having them as the same auditor as they have the same purpose: ensuring that the container is not running as root (either by overriding the UID or failing to run). Although it will increase the complexity of the auditor I could try that if everyone agree on this solution.

dani-santos-code:
Hi, @jcbbc! 👋 Thanks a lot for your contribution! 🙏🏽

I agree with @genevieveluyt  that `nonroot` and `runasuser` auditors are mutually exclusive. Another thing I noticed when testing locally was that `runAsNonRoot` seems to take precedence. I tried creating a pod with `runAsNonRoot` set to `true` and `runAsUser` set to 0. As a result, I got the following:

`Warning  Failed     9s (x4 over 36s)  kubelet            Error: container's runAsUser breaks non-root policy`

So, if I understand correctly the `nonroot` auditor already covers the security concern of not running the container as a root user? What are your thoughts?

jcbbc:
Hi @dani-santos-code 

I think both cover the same goal (having non-root containers) but use a different strategy (preventing images with root user to be run vs overriding UID). However as of today, when forcing the UID through the `runAsUser` field without setting the `runAsNonRoot` to `true`, the `nonroot` auditor will fail indicating that the container may run as root which is false as the `runAsUser` is there to enforce a non-root user. This makes `kubeaudit` unreliable when using the `runAsUser` strategy instead of the `runAsNonRoot` one. However both are valid strategies in my opinion.

But I agree that in the end it would be best to have only one auditor checking for both. If either `runAsUser` or `runAsNonRoot` is used, then we know that the container won't run as root. 

I'm working on a new version witch checks for `runAsUser` implemented into the `nonroot` auditor.

jcbbc:
I've updated the PR by merging `runasuser` into `nonroot`. It checks for `runAsUser` usage before `runAsNonRoot` as forcing a user makes `runAsNonRoot` useless.

For now I didn't add a `--user` to check and fix a specific user as it will make the whole thing more complex. Maybe once we're satisfied with the current merging. As a consequence, no automated fixing will be available when `runAsUser` is set to `0`

auditors/nonroot/nonroot.go
```diff
@@ -127,3 +162,15 @@ func isContainerRunAsNonRootCSCFalse(container *k8stypes.ContainerV1) bool {
 func isContainerRunAsNonRootNil(container *k8stypes.ContainerV1) bool {
 	return container.SecurityContext == nil || container.SecurityContext.RunAsNonRoot == nil
 }
+
+func isContainerRunAsUserNil(container *k8stypes.ContainerV1) bool {
+	return container.SecurityContext == nil || container.SecurityContext.RunAsUser == nil
+}
+
+func isPodRunAsUserNil(podSpec *k8stypes.PodSpecV1) bool {
+	if podSpec.SecurityContext == nil || podSpec.SecurityContext.RunAsUser == nil {
```
dani-santos-code, line 171: can we not do the same as you did for `isContainerRunAsUserNil` here?

`return podSpec.SecurityContext == nil || podSpec.SecurityContext.RunAsUser == nil`

auditors/nonroot/nonroot.go
```diff
@@ -59,16 +98,12 @@ func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource)
...
-	if podSpec == nil {
-		return nil
-	}
 	if isContainerRunAsNonRootNil(container) {
 		if isPodRunAsNonRootNil(podSpec) {
 			return &kubeaudit.AuditResult{
 				Name:     RunAsNonRootPSCNilCSCNil,
 				Severity: kubeaudit.Error,
-				Message:  "runAsNonRoot is not set in container SecurityContext nor the PodSecurityContext. It should be set to 'true' in at least one of the two.",
+				Message:  "neither runAsNonRoot nor runAsUser are set in container SecurityContext or in the PodSecurityContext. runAsNonRoot should be set to 'true' or runAsUser should be set to a non-root UID either in container SecurityContext or PodSecurityContext.",
```
dani-santos-code, line 106:
```suggestion
				Message:  "runAsNonRoot should be set to 'true' or runAsUser should be set to a non-root UID either in the container SecurityContext or PodSecurityContext.",
```

dani-santos-code: I think it would be a bit more succinct/clearer if we make the message a bit shorter. What do you think?

jcbbc: Indeed better. Done.

auditors/nonroot/nonroot.go
```diff
@@ -82,7 +117,7 @@ func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource)
 			return &kubeaudit.AuditResult{
 				Name:     RunAsNonRootPSCFalseCSCNil,
 				Severity: kubeaudit.Error,
-				Message:  "runAsNonRoot is not set in container SecurityContext and is set to false in the PodSecurityContext. It should be set to 'true' in at least one of the two.",
+				Message:  "runAsNonRoot is not set in container SecurityContext but is set to false in the PodSecurityContext while runAsUser is not set. runAsNonRoot should be set to 'true' or runAsUser should be set to a non-root UID either in container SecurityContext or PodSecurityContext.",
```
dani-santos-code, line 120:
```suggestion
				Message:  "runAsNonRoot is set to false in the PodSecurityContext. runAsNonRoot should be set to 'true' or runAsUser should be set to a non-root UID either in the container SecurityContext or PodSecurityContext.",
```

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,46 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+	if podSpec == nil {
+		return nil
+	}
+
+	if !isPodRunAsUserNil(podSpec) {
+		if *podSpec.SecurityContext.RunAsUser == 0 {
+			return &kubeaudit.AuditResult{
+				Name:     RunAsUserPSCRoot,
+				Severity: kubeaudit.Error,
+				Message:  "runAsUser is not set in container SecurityContext and is set to 0 in the PodSecurityContext. It should be set to > 0 in at least one of the two.",
```
dani-santos-code, line 77:
```suggestion
				Message:  "runAsUser is set to 0 in the PodSecurityContext. It should be set to a value > 0 in either the PodSecurityContext or in the container SecurityContext.",
```

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,46 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
 }
 
 func auditContainer(container *k8stypes.ContainerV1, resource k8stypes.Resource) *kubeaudit.AuditResult {
+	if !isContainerRunAsUserNil(container) {
+		if *container.SecurityContext.RunAsUser == 0 {
+			return &kubeaudit.AuditResult{
+				Name:     RunAsUserCSCRoot,
+				Severity: kubeaudit.Error,
+				Message:  "container running with root user (UID 0) as configured with runAsUser SecurityContext. It should be set to > 0.",
```
dani-santos-code, line 57:
```suggestion
				Message:  "runAsUser is set to UID 0 (root user) in the container SecurityContext. It should be set to a value > 0.",
```

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,46 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+
+		return nil
+	}
+
 	if isContainerRunAsNonRootCSCFalse(container) {
 		return &kubeaudit.AuditResult{
 			Name:     RunAsNonRootCSCFalse,
 			Severity: kubeaudit.Error,
-			Message:  "runAsNonRoot is set to false in container SecurityContext. It should be set to true.",
+			Message:  "runAsNonRoot is set to false and runAsUser is not set in container SecurityContext. Either the first should be set to true or the later set to a non-root UID.",
```
dani-santos-code, line 91:
```suggestion
			Message:  "runAsNonRoot is set to false in the container SecurityContext. It should be set to true. Another option is to set runAsUser to a non-root UID ( > 0)."
```

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,46 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+
+		return nil
+	}
+
 	if isContainerRunAsNonRootCSCFalse(container) {
 		return &kubeaudit.AuditResult{
 			Name:     RunAsNonRootCSCFalse,
 			Severity: kubeaudit.Error,
-			Message:  "runAsNonRoot is set to false in container SecurityContext. It should be set to true.",
+			Message:  "runAsNonRoot is set to false in the container SecurityContext. It should be set to true. Another option is to set runAsUser to a non-root UID ( > 0)."
```
dani-santos-code, line 91:
```suggestion
			Message:  "runAsNonRoot is set to false in the container SecurityContext. It should be set to true. Another option is to set runAsUser to a non-root UID ( > 0).",
```

dani-santos-code: I forgot to add the comma here when I made the suggestion. Sorry about that!

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,46 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+		}
+
+		return nil
+	}
+
+	podSpec := k8s.GetPodSpec(resource)
+	if podSpec == nil {
+		return nil
+	}
+
```
dani-santos-code, line 71: This is just an idea. As I was running the tests ("run-as-user-psc-0-csc-1.yml") I missed seeing some info/warning about the Pod Security Context with `runAsUser` set to 0. So I thought that we could change the code a bit so that we would get an error at an info level just to let the user know that runAsUser set to 0 is not recommended. If we change it, though, we will need to modify the tests a bit. Not sure it's worth it. What do you think?

```suggestion
	podSpec := k8s.GetPodSpec(resource)
	if podSpec == nil {
		return nil
	}

	if !isContainerRunAsUserNil(container) {
		if *container.SecurityContext.RunAsUser == 0 {
			return &kubeaudit.AuditResult{
				Name:     RunAsUserCSCRoot,
				Severity: kubeaudit.Error,
				Message:  "runAsUser is set to UID 0 (root user) in the container SecurityContext. It should be set to a value > 0.",
				Metadata: kubeaudit.Metadata{
					"Container": container.Name,
				},
			}
		} else {
			if !isPodRunAsUserNil(podSpec) {
				if *podSpec.SecurityContext.RunAsUser == 0 {
					return &kubeaudit.AuditResult{
						Name:     RunAsUserPSCRoot,
						Severity: kubeaudit.Info,
						Message:  "runAsUser is set to 0 in the PodSecurityContext. It should be set to a value > 0 in either the PodSecurityContext or in the container SecurityContext.",
						Metadata: kubeaudit.Metadata{
							"Container": container.Name,
						},
					}
				}
			}
		}
	}
```

dani-santos-code:
hey @jcbbc! This is looking really good! Thanks for making the changes and explaining your thoughts! I made some suggestions. Let me know what you think. We also need to update the `docs/auditors/nonroot.md` accordingly and add the logic to check and fix `runAsUser`. Let's see what others think as well! c/c @genevieveluyt

jcbbc:
Thanks! 

I've implemented your suggestion regarding the info message when PSC is set to 0, seems appropriate indeed. I've updated the tests accordingly.

The auditor doc has been [updated](https://github.com/Shopify/kubeaudit/pull/316/files#diff-cfa3620a7df0a4362f36dcc17eea57ff52f32c55b90ca2e7d7ace35450ab75e9) but tel me if you expect additional things.

Regarding the fix for `runAsUser`, I like your idea of a `--user` argument however, I'm wondering how it should be used. Does specifying an UID mean that we should check that this UID has been set for all `runAsUser` and so failed even if a `runAsUser` specified a non-root user but one that is different that the one specified with `--user`. Or should we only use `--user` for the fix part and not in the check part at all?

jcbbc:
So I've played with a `--user` arguments and here are my thoughts:
- using it only for fixing: I've find it weird to have an argument that is only used for fixing while every others auditors arguments are used for checks.
- using it both for checks and fixing: It makes the auditor fairly complex. It's now checking for `runAsNonRoot: false`, `runAsUser: 0`  and `runAsUser` not set to a given user and these both in the container security context and pod security context. And what if we pass `--user 0`? Should it be accepted as valid argument even if it go against the very purpose of the auditor?

The only alternative I see would be to not offer any auto-fix for `runAsUser: 0`

What's your opinion?

genevieveluyt:
Hey @jcbbc , thanks for making the changes! I will review the code this week, but to answer your question, I agree that it's unusual to use a flag only for fixing and not for auditing but at the same time I don't think it's useful to apply the flag to auditing. Personally, we always fix using `runAsNonRoot: true` so excluding the flag is not an issue for us, but if you would like support for auto-fixing by setting `runAsUser`, I agree we should require the flag argument to be greater than 0 (a negative value also doesn't make sense). Maybe the flag name can be more descriptive to indicate it's only used for fixing, eg. `fix-user`?

So I think the options are
- Don't support the flag and always fix by setting `runAsNonRoot: true`
- Name the flag `fix-user` (or something) and only allow values greater than 0
- Name the flag `fix-using-run-as-user` and have it be a boolean option which, when set to true, will fix by setting `runAsUser` to some default value we decide (probably `1`?)

I'm sure there are others that I can't think of at this moment 🙂

auditors/nonroot/fixtures/run-as-user-psc-0-allowed-multi-containers-single-label.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod
+  labels:
+    name: pod
+    container.audit.kubernetes.io/container1.allow-run-as-root: "SuperuserPrivilegesNeeded"
```
genevieveluyt, line 7: I'm a bit confused why setting an override label for the container results in `runAsUser: 0` being allowed on the pod... Is this the behaviour we want? Is it consistent with `runAsNonRoot`?

auditors/nonroot/fixtures/run-as-user-psc-1-run-as-non-root-psc-true.yml
```diff
@@ -0,0 +1,14 @@
...
+kind: Pod
+metadata:
+  name: pod
+  labels:
+    name: pod
+  namespace: run-as-user-psc-1-run-as-non-root-psc-true
+spec:
+  securityContext:
+    runAsUser: 1
+    runAsNonRoot: false
```
genevieveluyt, line 11:
```suggestion
    runAsNonRoot: true
```

genevieveluyt (commented):
Awesome, thanks for adding all the tests. Just some small things

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,59 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+					Severity: kubeaudit.Info,
+					Message:  "runAsUser is set to 0 in the PodSecurityContext. It should be set to a value > 0 in either the PodSecurityContext or in the container SecurityContext.",
+					Metadata: kubeaudit.Metadata{
+						"Container": container.Name,
+					},
+				}
+			}
+		}
+
+		return nil
```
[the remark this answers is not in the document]
jcbbc: This was a suggestion from @dani-santos-code to warn the user of a dangerous practice even if in practice, it doesn't affect the current container as the Pod Security Context is overridden by the container one. I though it was indeed a good idea. Notice that the severity is set to `info`  not `warning` or `error` so it shouldn't fail the audit.

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,59 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+	if podSpec == nil {
+		return nil
+	}
+
+	if !isContainerRunAsUserNil(container) {
+		if *container.SecurityContext.RunAsUser == 0 {
+			return &kubeaudit.AuditResult{
+				Name:     RunAsUserCSCRoot,
+				Severity: kubeaudit.Error,
+				Message:  "runAsUser is set to UID 0 (root user) in the container SecurityContext. It should be set to a value > 0.",
```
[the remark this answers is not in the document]
jcbbc: Will do that indeed.

auditors/nonroot/fixtures/run-as-user-psc-0-allowed-multi-containers-single-label.yml
jcbbc: Hum indeed, the override label should be for the pod or container2 if the goal is to allow them to run as root. container1 won't run as root anyway as it runs with user 1.

I will check that test.

auditors/nonroot/fixtures/run-as-user-psc-1-run-as-non-root-psc-true.yml
jcbbc: The goals of these tests (the one with `true` and this one with `false`) was to ensure that the `runAsRoot`, whatever its value is, doesn't have any incidence on the result if `runAsUser` is used. But maybe this test is overkill and we can delete it?

jcbbc:
> Hey @jcbbc , thanks for making the changes! I will review the code this week, but to answer your question, I agree that it's unusual to use a flag only for fixing and not for auditing but at the same time I don't think it's useful to apply the flag to auditing. Personally, we always fix using `runAsNonRoot: true` so excluding the flag is not an issue for us, but if you would like support for auto-fixing by setting `runAsUser`, I agree we should require the flag argument to be greater than 0 (a negative value also doesn't make sense). Maybe the flag name can be more descriptive to indicate it's only used for fixing, eg. `fix-user`?
> 
> So I think the options are
> 
> * Don't support the flag and always fix by setting `runAsNonRoot: true`
> * Name the flag `fix-user` (or something) and only allow values greater than 0
> * Name the flag `fix-using-run-as-user` and have it be a boolean option which, when set to true, will fix by setting `runAsUser` to some default value we decide (probably `1`?)
> 
> I'm sure there are others that I can't think of at this moment 🙂

On my side we won't use the auto-fixing as we have a practice to randomly generated UIDs for containers so they doesn't share the same UID on a the host. So only keeping auto-fixing for `runAsNonRoot` is fine for me. So my preference would be to not implement fixing for now and to keep that when someone has a real use case.

Otherwise, if we still want fixing, I would indeed go for the `--fix-user` option and not to use the specified UID for auditor. I've played with the implementation and using it in the auditor part really makes the thing convoluted.

auditors/nonroot/nonroot.go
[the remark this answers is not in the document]
genevieveluyt: Ah gotcha, I totally missed that they had different severities. Based on the descriptions of the different severity levels, I think this should be a warning though? https://github.com/Shopify/kubeaudit#audit-results

auditors/nonroot/fixtures/run-as-user-psc-1-run-as-non-root-psc-true.yml
genevieveluyt: They are both set to false currently. The name of this file has `run-as-non-root-psc-true` so I assumed it was intended to be `true` 😉

genevieveluyt:
Hey @jcbbc, just keeping fixing logic for `runAsNonRoot` (as it works currently) sounds good to me! As you said, if another use case comes up later we can revisit.

jcbbc:
Ok thanks for your feedback. I will try to take some time to finish this PR this week so we can merge it. Thanks!

jcbbc:
I've updated the PR with the last feedbacks. I think it's ready to merge now.

genevieveluyt:
Awesome thanks! I'll have a final look next week.

dani-santos-code:
Hey @jcbbc, this is looking great. Thank you!I tested locally and it works well. The only thing I miss is running autofix and getting rid of runAsUser in case it's set to 0. I was expecting to at least not see  `runAsUser:0`disappear and instead get `runAsNonRoot: true`.Currently, if I set it to 0 and run `autofix`, this is what I get:

this is the manifest (with `runAsUser:0`):
```
apiVersion: v1
kind: Pod
metadata:
  name: pod
  labels:
    name: pod
  namespace: test
spec:
  containers:
    - name: container
      image: scratch
      securityContext:
        runAsUser: 0
```

This is what I get after running `autofix`:

```
apiVersion: v1
kind: Pod
metadata:
  name: pod
  labels:
    name: pod
  namespace: test
  annotations:
    container.apparmor.security.beta.kubernetes.io/container: runtime/default
    seccomp.security.alpha.kubernetes.io/pod: runtime/default
spec:
  containers:
    - name: container
      image: scratch
      securityContext:
        runAsUser: 0
        capabilities:
          drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETPCAP
            - SETUID
            - SYS_CHROOT
        privileged: false
        readOnlyRootFilesystem: true
      resources: {}
  automountServiceAccountToken: false
```

As you can see, it keeps `runAsUser:0` and it doesn't add the `runAsNonRoot:true`

I was expecting the following (dropping runAsUser and setting `runAsNonRoot:true`) :

```
apiVersion: v1
kind: Pod
metadata:
  name: pod
  labels:
    name: pod
  namespace: test
  annotations:
    container.apparmor.security.beta.kubernetes.io/container: runtime/default
    seccomp.security.alpha.kubernetes.io/pod: runtime/default
spec:
  containers:
    - name: container
      image: scratch
      securityContext:
        capabilities:
          drop:
            - AUDIT_WRITE
            - CHOWN
            - DAC_OVERRIDE
            - FOWNER
            - FSETID
            - KILL
            - MKNOD
            - NET_BIND_SERVICE
            - NET_RAW
            - SETFCAP
            - SETGID
            - SETPCAP
            - SETUID
            - SYS_CHROOT
        privileged: false
        readOnlyRootFilesystem: true
        runAsNonRoot: true
      resources: {}
  automountServiceAccountToken: false
```

I believe you can achieve that by adding the following check in this function: https://github.com/Shopify/kubeaudit/blob/master/auditors/nonroot/fix.go#L18

```go
	if *f.container.SecurityContext.RunAsUser == 0 {
		f.container.SecurityContext.RunAsUser = nil
	}
```

And then add `PendingFix: &fixRunAsNonRoot{container: container}` in the `auditContainer` function	where you run checks like `f *podSpec.SecurityContext.RunAsUser == 0`

Let me know what you think! Other than that, it looks good to go! 🎉

jcbbc:
Yes, make sense and allow to have a consistent fixing strategy at least.

I've implemented your suggestion.

auditors/nonroot/nonroot.go
```diff
@@ -45,11 +49,65 @@ func (a *RunAsNonRoot) Audit(resource k8stypes.Resource, _ []k8stypes.Resource)
...
+	if podSpec == nil {
+		return nil
+	}
+
+	if !isContainerRunAsUserNil(container) {
+		if *container.SecurityContext.RunAsUser == 0 {
+			return &kubeaudit.AuditResult{
+				Name:     RunAsUserCSCRoot,
+				Severity: kubeaudit.Error,
+				Message:  "runAsUser is set to UID 0 (root user) in the container SecurityContext. Either set it to a value > 0 or set runAsNonRoot to false.",
```
genevieveluyt, line 62: Sorry I missed this the first time 🙈 . Just setting `runAsNonRoot: false` won't actually fix anything if `runAsUser: 0` is set.

```suggestion
				Message:  "runAsUser is set to UID 0 (root user) in the container SecurityContext. Either set it to a value > 0 or remove it and set runAsNonRoot to false.",
```

dani-santos-code: Another nit: `runAsNonRoot` should be  set to `true` so:

`"runAsUser is set to UID 0 (root user) in the container SecurityContext. Either set it to a value > 0 or remove it and set runAsNonRoot to true.",`

jcbbc: My bad, didn't saw that too. I've fixed all the messages.

genevieveluyt (approved):
LGTM!

dani-santos-code (approved):

jcbbc:
Thanks everyone! 

In order to keep you busy, I've already pushed a new PR :) Tell me what you think of the idea!
https://github.com/Shopify/kubeaudit/pull/322

PR #322, opened 2021-02-03
Adds a new 'mounts' command to audit sensitive host paths mounts

jcbbc:
##### Description

Adds a new command `mounts` to audit sensitive host paths that are mounted in contains. The auditor will trigger an alert when one of the specified paths is mounted by a container. If no paths are specified in arguments, a default paths list is used (I use the one used by [Falco](https://github.com/falcosecurity/falco/blob/master/rules/k8s_audit_rules.yaml#L155)).

Alerts can be overridden using a label either at the pod level or the container level. Labels are per mount, so you can only override alerts for legit mounts.

Here's a sample result:
```
kubeaudit mounts -f /Users/j.courtial/dev/go/src/github.com/kubeaudit/auditors/mounts/fixtures/proc-mounted.yml

  apiVersion: v1
  kind: Pod
  metadata:
    name: pod
    namespace: proc-mounted

-- [error] SensitivePathsMounted
   Message: Sensitive path mounted as volume: proc-volume (/proc -> /host/proc, readOnly: false). It should be removed from the container's mounts list.
   Metadata:
      Container: container
      Mount: proc-volume
```

This auditor makes the `mountds` kind of redundant so we can remove/deprecate it if you prefer. 

##### Type of change
- [x] New feature :sparkles:
- [x] This change requires a documentation update 📖

#####  How Has This Been Tested?
- [x] Automated tests
#####  Checklist:
- [x] I have 🎩 my changes (A 🎩 specifically includes pulling down changes, setting them up, and manually testing the changed features and potential side effects to make sure nothing is broken)
- [x] I have signed the appropriate Contributor License Agreement


genevieveluyt:
Hey @jcbbc , thanks for opening another PR! The team will triage this on Monday. Cheers!

docs/auditors/mounts.md
```diff
@@ -0,0 +1,139 @@
...
+| /var/lib/kubelet          |  Directory for Kublet-related configuration                                                             |
+| /var/lib/kubelet/pki      |  Directory containing the certificate and private key of the Kublet                                                               |
+| /etc/kubernetes           |  Directory containing Kubernetes related configuration              |
+| /etc/kubernetes/manifests |  Directory containing manifest of Kubernetes components                                                         |
+
+
+## Examples
+
+```
+$ kubeaudit mounts -f /Users/j.courtial/dev/go/src/github.com/kubeaudit/auditors/mounts/fixtures/proc-mounted.yml
```
genevieveluyt, line 39:
```suggestion
$ kubeaudit mounts -f auditors/mounts/fixtures/proc-mounted.yml
```

docs/auditors/mounts.md
```diff
@@ -0,0 +1,139 @@
...
+        - name: proc-volume
+          hostPath:
+            path: /proc
+```
+
+## Override Errors
+
+First, see the [Introduction to Override Errors](/README.md#override-errors).
+
+The override identifier has the format `allow-host-path-mount--[mount name]` which allows for each mount to be individually overridden.
```
genevieveluyt, line 88: I think the double dash should be a single dash to match the code and examples. Honestly, I kind of like the double dash better but single dash would be consistent with capabilities overrides... What do you think?

Also, I saw some logic to convert the volume name to override label (lowercase and replace underscores with dashes). Maybe you did that to be consistent with the capabilities auditor, but if the character set for volume names is allowed in labels I wouldn't be opposed to just using the volume name in the override label without any conversion. For example if the volume name is `my_Volume` then the override label would be `allow-host-path-mount--my_Volume`. What do you think? (I don't actually know if `my_Volume` is a valid volume name but just for argument sake 😄 )

auditors/mounts/config.go
```diff
@@ -0,0 +1,12 @@
+package mounts
+
+type Config struct {
+	SensitivePaths []string `yaml:"paths"`
+}
+
+func (config *Config) GetSensitivePaths() []string {
+	if config == nil {
+		return nil
```
genevieveluyt, line 9: Can we return the default?

genevieveluyt (commented):
This is really awesome, thank you! Much better implemented than the `mountds` auditor too 😄 

I think you're right and we should deprecate `mountds` in favour of this one. There's not really a proper deprecation flow but I think we can just remove it from the `all` command and update the docs to say it's deprecated.

Could you also update the kubeaudit config docs in the [main README](https://github.com/Shopify/kubeaudit#configuration-file) and add examples using the config file and flag ([for example](https://github.com/Shopify/kubeaudit/blob/master/docs/auditors/capabilities.md#example-with-config-file))

Would you mind modifying one of the tests (or adding a test) to test at least one other host path (other than `/proc`) to prove it doesn't only work for the first path in the array.

docs/auditors/mounts.md
jcbbc: Totally right for both. The double dash was a typo. Regarding the volume name, indeed let's keep things simple and use the volume name as it is 👍

auditors/mounts/mounts.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var auditResults []*kubeaudit.AuditResult
+
+	for _, mount := range container.VolumeMounts {
+		if _, ok := sensitiveVolumes[mount.Name]; ok {
+			volume := sensitiveVolumes[mount.Name]
+				auditResults = append(auditResults, &kubeaudit.AuditResult{
+					Name:     SensitivePathsMounted,
+					Severity: kubeaudit.Error,
+					Message:  fmt.Sprintf("Sensitive path mounted as volume: %s (%s -> %s, readOnly: %t). It should be removed from the container's mounts list.", mount.Name, volume.HostPath.Path, mount.MountPath, mount.ReadOnly),
```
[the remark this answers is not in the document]
jcbbc: I find both to be valid :) My reasoning in putting the host path first was that this auditor is looking for sensitive host paths and the message is "Sensitive path mounted as volume" so I though it would make sense to display the sensitive path first. But I'm totally ok to swap if you think it's indeed better :)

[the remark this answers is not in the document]
genevieveluyt: The arrow makes me think it's directional, and having the mount path point to the host path makes more sense to me, but if even we aren't sure then this is probably going to cause confusion and I would rather just omit the mount path (it should probably be part of the result metadata anyway). The host path is the important part and the mount name should be enough to identify where the problem is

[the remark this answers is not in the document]
jcbbc: Ok, will do that 👌  What abut the `readOnly` info then? If the `mountPath` is no more in the main message, I'm not sure it makes sense to have it there too. Should we put it in the metadata too or just remove it altogether?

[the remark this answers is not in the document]
genevieveluyt: Yea good point, probably remove the `readOnly` from the message as well. I don't think it hurts to include in the metadata if you think it's useful information to know when reviewing results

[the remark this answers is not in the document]
jcbbc: My idea for displaying this value was that you may want to put extra care in checking container that can write on the host because that a big step further in terms of privileges. As you said it won't hurt to have it in the Metadata 🤷

cmd/commands/all.go
```diff
@@ -47,6 +47,10 @@ func setConfigFromFlags(cmd *cobra.Command, conf config.KubeauditConfig) config.
 		conf.AuditorConfig.Capabilities.AllowAddList = capabilitiesConfig.AllowAddList
 	}
 
+	if flagset.Changed("paths") {
```
genevieveluyt, line 50: Good catch, I forgot about this 😂  Can we make `paths` a constant though like `capsAddFlagName`?

auditors/mounts/mounts_test.go
```diff
@@ -0,0 +1,34 @@
...
+		expectedErrors []string
+	}{
+		{"docker-sock-mounted.yml", fixtureDir, []string{SensitivePathsMounted}},
+		{"proc-mounted.yml", fixtureDir, []string{SensitivePathsMounted}},
+		{"proc-mounted-allowed.yml", fixtureDir, []string{override.GetOverriddenResultName(SensitivePathsMounted)}},
+		{"proc-mounted-allowed-multi-containers-multi-labels.yml", fixtureDir, []string{override.GetOverriddenResultName(SensitivePathsMounted)}},
+		{"proc-mounted-allowed-multi-containers-single-label.yml", fixtureDir, []string{SensitivePathsMounted, override.GetOverriddenResultName(SensitivePathsMounted)}},
+	}
+
+	config := Config{SensitivePaths: DefaultSensitivePaths}
```
genevieveluyt, line 26: I think we can let this fallback to the default in the config now?
```suggestion
	config := Config{}
```

cmd/commands/all.go
genevieveluyt, line 50: Good catch, I forgot about this! Can we make `"paths"` into a constant though like `capsAddFlagName`?

docs/auditors/mounts.md
```diff
@@ -0,0 +1,248 @@
...
+  - name: etc-volume
+    hostPath:
+      path: /etc
+  - name: docker-socket-volume
+    hostPath:
+      path: /var/run/docker.sock
+```
+
+```shell
+$ kubeaudit capabilities --path "/etc,/var/run/docker.sock" -f "manifest.yaml"
```
genevieveluyt, line 143:
```suggestion
$ kubeaudit capabilities --paths "/etc,/var/run/docker.sock" -f "manifest.yaml"
```

genevieveluyt (commented):
Docs are looking good!

docs/auditors/mounts.md
jcbbc: Good catch. Also the `capabilities` instead of `mounts` ... Copy/paste 🙄

jcbbc:
Should we address the deprecation in this PR or another one?

auditors/mounts/mounts.go
```diff
@@ -0,0 +1,120 @@
...
+		if volume, ok := sensitiveVolumes[mount.Name]; ok {
+			auditResults = append(auditResults, &kubeaudit.AuditResult{
+				Name:     SensitivePathsMounted,
+				Severity: kubeaudit.Error,
+				Message:  fmt.Sprintf("Sensitive path mounted as volume: %s (hostPath: %s). It should be removed from the container's mounts list.", mount.Name, volume.HostPath.Path),
+				Metadata: kubeaudit.Metadata{
+					"Container":              container.Name,
+					MountNameMetadataKey:     mount.Name,
+					MountPathMetadataKey:     mount.MountPath,
+					MountReadOnlyMetadataKey: fmt.Sprintf("%t", mount.ReadOnly),
```
genevieveluyt, line 109: Actually can we add the volume name and host path as metadata as well? It might be useful for JSON output

docs/auditors/mounts.md
```diff
@@ -0,0 +1,253 @@
...
+    name: deployment
+    namespace: example-namespace
+
+--------------------------------------------
+
+-- [error] SensitivePathsMounted
+   Message: Sensitive path mounted as volume: etc-volume (/etc -> /host/etc, readOnly: false). It should be removed from the container's mounts list.
+   Metadata:
+      Container: container
+      Mount: etc-volume
```
genevieveluyt, line 117: Do you mind adding the other metadata fields here?

genevieveluyt (commented):
Just a few small things and then I think it should be good to merge! Let's do the deprecation separately.

genevieveluyt (approved):
Awesome, thanks so much!

docs/auditors/mounts.md
```diff
@@ -0,0 +1,273 @@
...
+| :------------------------ | :--------------------------------------------------------------------------------------- |
+| /proc                     |  Pseudo-filesystem which provides an interface to kernel data structures                                                  |
+| /var/run/docker.sock      |  Unix socket used to communicate with Docker daemon                                   |
+| /                         |  Filesystem's root |
+| /etc                      |  Directory that usually contains all system related configurations files         |
+| /root                     |  Home directory of the `root` user                                             |
+| /var/run/crio/crio.sock   |  Unix socket used to communicate with the CRI-O Container Engine                                                 |
+| /home/admin               |  Home directory of the `admin` user        |
+| /var/lib/kubelet          |  Directory for Kublet-related configuration                                                             |
+| /var/lib/kubelet/pki      |  Directory containing the certificate and private key of the Kublet                                                               |
```
dani-santos-code, line 31: tiny thing:
```suggestion
| /var/lib/kubelet/pki      |  Directory containing the certificate and private key of the kubelet                                                               |
```

docs/auditors/mounts.md
```diff
@@ -0,0 +1,273 @@
...
+        - name: etc-volume
+          hostPath:
+            path: /etc
+        - name: docker-socket-volume
+          hostPath:
+            path: /var/run/docker.sock
+```
+
+```shell
+$ kubeaudit mounts --kconfig "config.yaml" -f "manifest.yaml"
```
dani-santos-code, line 105: Passing in a config file only works with `all` and `autofix`(https://github.com/Shopify/kubeaudit#configuration-file)

```suggestion
$ kubeaudit all --kconfig  "config.yaml" -f "manifest.yaml"
```

docs/auditors/mounts.md
```diff
@@ -0,0 +1,273 @@
...
+paths you'd like to have kubeaudit raise an error when they are mounted in a container.
+
+`config.yaml`
+
+```yaml
+---
+auditors:
+  mounts:
+    paths: ["/etc", "/var/run/docker.sock"]
+```
```
dani-santos-code, line 74: When I tested locally, I also had to turn this auditor on by setting `mounts`to true, like so:

```yaml
enabledAuditors:
  mounts: true
 ```

Comment (dani-santos-code) [reply docs/auditors/mounts.md:74]: Another suggestion, to make it more explicit (completely optional). Instead of paths, how about `pathsToWatch`?

File: docs/auditors/mounts.md [hunk 13]
```diff
@@ -0,0 +1,273 @@
...
+      MountVolume: proc-volume
+      MountVolumeHostPath: /proc
+
+```
+
+### Example with Config File
+
+A custom paths list can be provided in the config file. See [docs](docs/all.md) for more information. These are the host
+paths you'd like to have kubeaudit raise an error when they are mounted in a container.
+
```
dani-santos-code, line 66:
```suggestion
If you don't want kubeaudit to raise errors for all the paths in the default list (`DefaultSensitivePaths`), you can provide a custom paths list in the config file. See [docs](docs/all.md) for more information. That way kubeaudit will only raise errors for those specific paths listed in the config file.
```

dani-santos-code: something along these lines... Because what happens here is that it overrides the default list, so it will only wtach the paths I specify

cmd/commands/mounts.go
```diff
@@ -0,0 +1,48 @@
...
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(mountsCmd)
+	setPathsFlags(mountsCmd)
+}
+
+func setPathsFlags(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVarP(&mountsConfig.SensitivePaths, sensitivePathsFlagName, "s", mounts.DefaultSensitivePaths,
```
dani-santos-code, line 37: Another tiny thing, and more of a question. How about `p` rather than `s` for the shorthand?
```suggestion
	cmd.Flags().StringSliceVarP(&mountsConfig.SensitivePaths, sensitivePathsFlagName, "p", mounts.DefaultSensitivePaths,
```

dani-santos-code:
This is looking great, @jcbbc! I agree with @genevieveluyt that this a more comprehensive implementation, if compared to our current `mountds` auditor! Thank you!!! I left some comments/questions. So it makes sense to deprecate/remove the old auditor. 🎉🎉🎉

cmd/commands/mounts.go
jcbbc: Yes, that was my initial idea but `-p` is already used as a shorthand for the `--format` argument (I guess it comes from `pretty`?)
```
  -p, --format string        The output format to use (one of "pretty", "logrus", "json") (default "pretty")
```

So I fall back on `s`  for "**S**ensitive paths" but it's indeed not really obvious

docs/auditors/mounts.md
jcbbc: I went for the simple `paths` to be more aligned with other auditors that also use simple key (`add`, `image`, `cpu`)  that are also the same as the CLI options. But we can totally switch to `pathsToWatch` if you find it better.

dani-santos-code: I see. For the capabilities auditor, however, we add capabilities to an allowAddList  so `kubeaudit` won't raise errors for those.  So it made me think that paths would have a similar effect - listed paths would be the ones `kubeaudit` would ignore. However it's the opposite - `kubeaudit` will only watch for those specific paths, everything else will be ignored. That's why I thought of a more explicit name... what do you think?

cmd/commands/mounts.go
dani-santos-code: oh, yeah! you're right! perhaps `sp`? not sure...it's optional. We can always revisit that later if someone gives use feedback saying it's confusing.

docs/auditors/mounts.md
jcbbc: Ok understood. I find the phrasing a bit weird however :) I was about to suggest `sensitivesPaths` but I guess the issue is the same, it may not be quite clear that it work as a deny-list and not as an allow-list. So if you prefer, yes we can go for `pathsToWatch`

dani-santos-code: yeah... let's leave it like then, then. And hopefully the documentation will make it clear! 😁

dani-santos-code (dismissed):

docs/auditors/mounts.md
dani-santos-code: or maybe `denyPathsList`? :P

dani-santos-code: @jcbbc can you please change it to `denyPathsList`? with that, we're good to go! thanks! 🙏

jcbbc: Sorry, I missed your previous comment. I've made the change.

dani-santos-code: Thanks a lot, @jcbbc!! :)

docs/auditors/mounts.md
```diff
@@ -10,9 +10,9 @@ kubeaudit mounts [flags]
 
 ### Flags
 
-| Short   | Long      | Description                                                          | Default                                                                  |
-| :------ | :-------- | :------------------------------------------------------------------- | :----------------------------------------------------------------------- |
-| -s      | --paths   | List of sensitive paths that shouldn't be mounted.                   | [default sensitive host paths list](#Default-sensitive-host-paths-list)  |
+| Short   | Long              | Description                                                          | Default                                                                  |
+| :------ | :---------------- | :------------------------------------------------------------------- | :----------------------------------------------------------------------- |
+| -s      | --denyPathsList   | List of sensitive paths that shouldn't be mounted.                   | [default sensitive host paths list](#Default-sensitive-host-paths-list)  |
```
dani-santos-code, line 15: I believe we can go for `deny-paths-list` so the flag has the same format as [the others](https://github.com/Shopify/kubeaudit/blob/8be8c73b59a5cac5d1e5178d18f721838ba6b119/cmd/commands/capabilities.go#L12)

docs/auditors/mounts.md
```diff
@@ -154,7 +154,7 @@ volumes:
 ```
 
 ```shell
-$ kubeaudit mounts --paths "/etc,/var/run/docker.sock" -f "manifest.yaml"
+$ kubeaudit mounts --denyPathsList "/etc,/var/run/docker.sock" -f "manifest.yaml"
```
dani-santos-code, line 157: if we change the flag, then this will be:

```$ kubeaudit mounts --deny-paths-list "/etc,/var/run/docker.sock" -f "manifest.yaml"```

docs/auditors/mounts.md
jcbbc: Oh yes good catch. Also, I was wondering, do we replace `-s` by `-d` for the `deny` part of the argument?

dani-santos-code: I guess it'd be a good idea to have `-d` as a shorthand, @jcbbc!

jcbbc:
Should be good now 👌

dani-santos-code:
Thank you @jcbbc !!!

dani-santos-code (approved):

PR #330, opened 2021-03-23
Removes mountds in favour of mounts

### LINKED ISSUE #328: Deprecate mountds
##### ISSUE TYPE
 - [ ] Bug Report
 - [ ] Feature Idea
- [x] Other
##### SUMMARY
With the new `mounts` auditor implementation, we no longer need `mountds` as discussed here: https://github.com/Shopify/kubeaudit/pull/322

It was decided that we should remove `mountds` in favour of `mounts`

This is required so that we can make a new release (0.13.0)

c/c @genevieveluyt @jcbbc

dani-santos-code:
##### Description

This PR removes `mountds` in favor of `mounts` which is an enhancement, since it checks not only for `/var/run/docker.sock`, but also other paths that are mounted, much like in[ Falco](https://github.com/falcosecurity/falco/blob/master/rules/k8s_audit_rules.yaml#L157). 

For more context, see https://github.com/Shopify/kubeaudit/pull/322

After this is merged, we can then release a new version of kubeaudit. 
Fixes #328

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [ ] Test A
- [ ] Test B

##### Checklist:


genevieveluyt (commented):
I think we should remove it from the `all` command but keep the `mountds` command separately and add a deprecation notice to the docs and maybe even the terminal output if it's set to `pretty` format. Something like "mountds is deprecated and will stop working in the next version. Please use mounts instead." Wdyt?

dani-santos-code:
Oh, interesting point. I thought we just wanted to remove the command. But the idea of sending a warning first about the deprecation is definitely more subtle/considerate for the users. Thanks! 🙏🏽

PR #332, opened 2021-04-15
Deprecate mountds command

genevieveluyt:
Removes `mountds` from the `all` command (should be a no-op since the `mounts` command tests for the same things and more).

Also adds a deprecation notice at the top of the results (if using pretty print output)

> DEPRECATION NOTICE: The 'mountds' command is deprecated and will stop working in a future minor release. Please use the 'mounts' command instead. If you use 'all' no change is required


dani-santos-code (approved):
I guess you need to remove `mountds` from the [config fixture](https://github.com/Shopify/kubeaudit/blob/master/config/config.yaml#L11), since config is used with `all` and we're removing it from there

dani-santos-code (approved):

PR #334, opened 2021-04-21
Version

### LINKED ISSUE #320: Add option to get kubeaudit version
##### ISSUE TYPE
 - [x] Feature Idea
# BUG REPORT
##### SUMMARY
Add option to get kubeaudit version

##### ENVIRONMENT
* Kubeaudit version:kubeaudit-v0.11.8
* Kubeaudit install method: brew and exe for windows

##### STEPS TO REPRODUCE

brew -v
brew --help

##### EXPECTED RESULTS

The current version could be shown

e.g
```
# kubeaudit -v
0.11.8
```

##### ACTUAL RESULTS

I found no way to currently retrieve the version, if there is a way, sry for bringing up this issue :) and i woul kindly ask to show me that way

##### ADDITIONAL INFORMATION

# FEATURE IDEA

maybe add a cmd line option (-v --version) to get the version or output it  on --help

genevieveluyt:
##### Description

Adds a version command which returns the current kubeaudit version.
```
$ kubeaudit version
0.14.0
```

Fixes #320

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

- [ ] Test A
- [ ] Test B

##### Checklist:


VERSION
```diff
@@ -0,0 +1 @@
+cmd/commands/VERSION
```
dani-santos-code, line 1: why do we need to symlink here?
Do we want to then do something like:
`export VERSION=$(cat VERSION )`

So then when we're describing release steps we can do something like:
`git tag -a $VERSION`?

dani-santos-code (approved):
really nice use of go embed! 🎉

VERSION
genevieveluyt: I thought it would just be easier to find the file. I'm pretty sure conventionally the VERSION file is in the root of the project but unfortunately go embed only lets you embed files in the same directory or a subdirectory (and it doesn't follow symlinks) so the real VERSION file has to be in cmd/commands. Editors like vscode let you edit the symlinked file as if it were the real file though.

dani-santos-code: oh, I see! Okie, then! Thanks for clarifying!

PR #335, opened 2021-04-22
Update release instructions

genevieveluyt:
It seems that goreleaser will now push up docker images even if set to draft mode.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,7 +13,6 @@ dockers:
   image_templates:
   - "shopify/kubeaudit:latest"
   - "shopify/kubeaudit:{{ .Tag }}"
-  - "shopify/kubeaudit:v{{ .Major }}"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -13,7 +13,6 @@ dockers:
   image_templates:
   - "shopify/kubeaudit:latest"
   - "shopify/kubeaudit:{{ .Tag }}"
-  - "shopify/kubeaudit:v{{ .Major }}"
```
dani-santos-code, line 16: so, by removing this line we guarantee that no breaking changes will be pushed automatically? Does that mean we will have to manually push those?

dani-santos-code (approved):

.goreleaser.yml
genevieveluyt: Since we're on major version v0 it doesn't really make sense to have that image at all (you might as well use latest). The release instructions currently just don't say to push it so we've never had it available on dockerhub.

Issue #338, opened 2021-05-18
Docker Images doesn't build correctly anymore


J12934:
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

# BUG REPORT

##### SUMMARY

##### ENVIRONMENT
* Kubeaudit version: 0.14.0
* Kubeaudit install method: Docker

##### STEPS TO REPRODUCE

```bash
git clone https://github.com/Shopify/kubeaudit.git
cd kubeaudit
docker build -t kubeaudit .
```

##### EXPECTED RESULTS

The docker build doesn't work correctly anymore as the image still uses golang 1.15 which isn't able to compile kubeaudit anymore since the use of embed #334.

##### ACTUAL RESULTS

```txt
[+] Building 35.2s (13/17)
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                  0.0s
 => => transferring dockerfile: 1.36kB                                                                                                                                                                                                                0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                     0.0s
 => => transferring context: 86B                                                                                                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/golang:1.15.1                                                                                                                                                                                      1.3s
 => [auth] library/golang:pull token for registry-1.docker.io                                                                                                                                                                                         0.0s
 => CACHED [builder 1/8] FROM docker.io/library/golang:1.15.1@sha256:bd7a499445abe17605c3c29ac4d924207c31f0274bcc011e933533ac603645e7                                                                                                                 0.0s
 => [internal] load build context                                                                                                                                                                                                                     0.1s
 => => transferring context: 450.37kB                                                                                                                                                                                                                 0.1s
 => [builder 2/8] RUN apt-get install --yes --no-install-recommends ca-certificates tzdata                                                                                                                                                            0.5s
 => [builder 3/8] RUN adduser --shell /bin/true --uid 1000 --disabled-login --no-create-home --gecos '' app   && sed -i -r "/^(app|root)/!d" /etc/group /etc/passwd   && sed -i -r 's#^(.*):[^:]*$#\1:/sbin/nologin#' /etc/passwd                     0.5s
 => [builder 4/8] WORKDIR /go/src/app/                                                                                                                                                                                                                0.0s
 => [builder 5/8] COPY go.mod go.sum ./                                                                                                                                                                                                               0.0s
 => [builder 6/8] RUN go mod download                                                                                                                                                                                                                30.4s
[...]

PR #339, opened 2021-05-18
🐛  Update dockerfile

genevieveluyt:
##### Description

As a result of recent changes which require Go 1.16 (namely, embed), the Dockerfile no longer works because it's using an older version of Go. Go release builds the binary locally and then copies it into the image so we didn't notice it before.

Fixes https://github.com/Shopify/kubeaudit/issues/338

##### Type of change

- [x] Bug fix :bug:
- [ ] New feature :sparkles:

- [x] `docker build -t kubeaudit .`

##### Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.15.1 AS builder
+FROM golang:1.16 AS builder
 
 # no need to include cgo bindings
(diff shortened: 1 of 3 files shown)


dani-santos-code (approved):
thanks for the quick fix!

PR #364, opened 2021-10-18
Remove wip probot

genevieveluyt:
The wip probot check is blocking dependabot PRs from passing CI... 

image

Since we don't have time to debug this right now, it's probably more desirable to get dependabot PRs shipped. I also removed the branch protection that required the check to pass.


dani-santos-code:
oh. interesting! I didn't notice the wip probot before!

PR #374, opened 2021-11-05
✨ Adds ability to not filter CRDS, see ISSUE #373

### LINKED ISSUE #373: support for CRD's that extend common types
##### ISSUE TYPE
 - [ ] Bug Report
 - [X] Feature Idea

##### SUMMARY
I've noticed that Kubeaudit doesn't catch CRDS that are extending typical k8s resources, like apps/v1.
It looks like either with finagling the data out with kubectl and scanning you can still get to the original source of truth.

##### STEPS TO REPRODUCE
create a custom CRD based off apps
scan k8s
don't find anything wrong with deploy applications

##### EXPECTED RESULTS
It should find them

##### other stuff

if you guys want to point me in the direction of where in the code one can specify a custom CRD, that'll work for me and I'll add a flag to support it.

nobletrout:
##### Description
Adds new command flag to not filter out generated resources, which unfortunately also can include CRD based resources (like a CRD based on `Application`)
Fixes #373 

##### Type of change

- [x]  New feature :sparkles:
I ran it against my cluster and compared results with the flag on and off and found  a lot more issues with the flag, which was my intent.

##### Checklist:

It is kind of difficult to test if these changes work or not, since they only apply to querying remote clusters and not local files. If you guys have a suggestion for me to add test cases, I'm happy to help or learn.


internal/k8sinternal/client.go
```diff
@@ -85,6 +85,7 @@ func IsRunningInCluster(client Client) bool {
 type ClientOptions struct {
 	// Namespace filters resources by namespace. Defaults to all namespaces.
 	Namespace string
+	IncludeGenerated bool
```
genevieveluyt, line 88: Could you add a comment here explaining what the field does so it'll end up in the Go docs?

cmd/commands/root.go
```diff
@@ -23,6 +23,7 @@ type rootFlags struct {
 	namespace   string
 	minSeverity string
 	exitCode    int
+	includegenerated bool
```
genevieveluyt, line 26: Can we make this camel case to be consistent?
```suggestion
	includeGenerated bool
```

genevieveluyt (commented):
Thanks for your PR! Could you please document the new flag in the [README](https://github.com/Shopify/kubeaudit/blob/main/README.md#global-flags). Perhaps something like:

"By default, kubeaudit will ignore generated resources (such as Pods generated by Deployments). If you would like kubeaudit to produce results for generated resources (for example if you have custom resources or want to catch orphaned resources where the owner resource no longer exists) you can use this flag."

genevieveluyt:
Perfect thanks! As for testing, I can look into adding a test next week.

If you'd like to try it out yourself, I'm thinking of doing something similar to https://github.com/Shopify/kubeaudit/blob/main/internal/test/test.go#L111-L114 with a new test fixture and then seeing if the report results contains the pods or not (or even just comparing the number of results). You can use [`make test-setup`](https://github.com/Shopify/kubeaudit/blob/main/Makefile#L39) to set up a test cluster using kind so you can run the individual test.

But definitely don't feel like you need to do all that, as I said I'm happy to add the testing portion.

nobletrout:
I'll try to get the unit tests to work, but if you beat me to it go ahead. I'm trying to figure out how to not run ALL the unit tests everytime.
any advice?

genevieveluyt (approved):
Ok I wrote out a test locally and it passes 🎉 . Since I can't push to your branch let's get your PR merged and I'll add the test as a separate PR.

PR #376, opened 2021-11-14
pkg/k8s: add support for initContainer

### LINKED ISSUE #355: kubeaudit doesn't support initContainers
##### ISSUE TYPE
 - [x] Bug Report
 - [ ] Feature Idea

# BUG REPORT

##### SUMMARY

As pointed out by @thriqon in https://github.com/Shopify/kubeaudit/issues/285#issuecomment-891289275, kubeaudit audits `containers` but not `initContainers`. This is specifically problematic for the `apparmor` auditor which warns if an apparmor annotation refers to a container that doesn't exist and will remove that annotation in `autofix`. Since it ignores `initContainers` this would mean that an apparmor annotation which correctly refers to an `initContainer` would be removed by kubeaudit. In general though, we should add support for kubeaudit to add security controls to `initContainers` the same way it does for `containers`.

##### ENVIRONMENT
See https://github.com/Shopify/kubeaudit/issues/285#issuecomment-891894619

##### STEPS TO REPRODUCE

See https://github.com/Shopify/kubeaudit/issues/285#issuecomment-891894619

##### EXPECTED RESULTS

Kubeaudit should audit `initContainers` the same way it audits `containers`.

##### ACTUAL RESULTS

Kubeaudit ignores `initContainers` and removes apparmor annotations that refer to `initContainers`

##### ADDITIONAL INFORMATION

See https://github.com/Shopify/kubeaudit/issues/285#issuecomment-891894619

danishprakash:
##### Description

Adds support for `initContainers`. Current state is as follows and as described [here](https://github.com/Shopify/kubeaudit/issues/355#issue-959191623):

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      annotations:
        container.apparmor.security.beta.kubernetes.io/nginx: runtime/default
        container.apparmor.security.beta.kubernetes.io/bb: runtime/default
    spec:
      initContainers:
      - image: busybox
        name: bb
        command:
          - /bin/echo
          - hello
      containers:
      - image: nginx
        name: nginx
```

```
$ kubeaudit apparmor -f deploy.yaml 

  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: nginx

-- [error] AppArmorInvalidAnnotation
   Message: AppArmor annotation key refers to a container that doesn't exist. Remove the annotation 'container.apparmor.security.beta.kubernetes.io/bb: runtime/default'.
   Metadata:
      Container: bb
      Annotation: container.apparmor.security.beta.kubernetes.io/bb: runtime/default

```

Fixes #355 

##### Type of change

- [x] Bug fix :bug:
- [ ] New feature :sparkles:

The tests are failing on the current main with the following error:
```
Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply
namespace/pod configured
Error from server (Forbidden): error when creating "../../internal/test/fixtures/all_resources/pod.yml": pods "pod" is forbidden: error looking up service account pod/default: serviceaccount "default" not found
```
Although I've tested my changes as part of this too by both removing the change (while keeping the fixtures) and vice versa. I'll take some time to look into the above error though.

##### Checklist:

### DIFF
--- a/auditors/apparmor/apparmor_test.go
+++ b/auditors/apparmor/apparmor_test.go
@@ -19,6 +19,8 @@ func TestAuditAppArmor(t *testing.T) {
 	}{
 		{"apparmor-enabled.yml", nil, true},
 		{"apparmor-annotation-missing.yml", []string{AppArmorAnnotationMissing}, true},
+		{"apparmor-annotation-init-container-enabled.yml", nil, true},
+		{"apparmor-annotation-init-container-missing.yml", []string{AppArmorAnnotationMissing}, true},
 		// These are invalid manifests so we should only test it in manifest mode as kubernetes will fail to apply it
 		{"apparmor-disabled.yml", []string{AppArmorDisabled}, false},
 		{"apparmor-invalid-annotation.yml", []string{AppArmorInvalidAnnotation}, false},
--- a/auditors/apparmor/fixtures/apparmor-annotation-init-container-enabled.yml
+++ b/auditors/apparmor/fixtures/apparmor-annotation-init-container-enabled.yml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod
+  namespace: apparmor-annotation-init-container-enabled
+  annotations:
+    container.apparmor.security.beta.kubernetes.io/container: localhost/someval
+    container.apparmor.security.beta.kubernetes.io/init-container: localhost/someval
+spec:
+  initContainers:
+  - name: init-container
+    image: scratch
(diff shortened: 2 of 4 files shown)


genevieveluyt (commented):
Nice! This is a very clean solution. Do you mind adding one more test to make sure kubeaudit produces an error if there is an init container without an apparmor annotation? (ie. a copy of the test fixture you added but remove line 8)

PR #381, opened 2021-11-19
Test IncludeGenerated

genevieveluyt:
##### Description

Adds a test for https://github.com/Shopify/kubeaudit/pull/374

Additionally:
- The service.yml `ClusterIP` was invalid ("The Service "test-service" is invalid: spec.clusterIP: Invalid value: "10.0.171.239": provided IP is not in the valid range. The range of valid IPs is 10.96.0.0/16")
- Importing the `test` package in `client_test.go` caused a circular dependency so updated the package from `k8sinternal` to `k8sinternal_test`

##### Type of change

- [ ] Bug fix :bug:
- [ ] New feature :sparkles:

- [ ] Test A
- [ ] Test B

##### Checklist:

### DIFF
--- a/internal/test/fixtures/service.yml
+++ b/internal/test/fixtures/service.yml
@@ -9,5 +9,5 @@ spec:
     - protocol: TCP
       port: 80
       targetPort: 9376
-  clusterIP: 10.0.171.239
+  clusterIP: 10.96.0.1
   type: LoadBalancer


internal/test/fixtures/service.yml
```diff
@@ -9,5 +9,5 @@ spec:
     - protocol: TCP
       port: 80
       targetPort: 9376
-  clusterIP: 10.0.171.239
-  type: LoadBalancer
\ No newline at end of file
+  clusterIP: 10.96.0.1
```
genevieveluyt, line 12: The service.yml ClusterIP was invalid ("The Service "test-service" is invalid: spec.clusterIP: Invalid value: "10.0.171.239": provided IP is not in the valid range. The range of valid IPs is 10.96.0.0/16").

This doesn't affect tests or anything. I just happened to try to `kubectl apply` the service into a kind cluster and it failed.

internal/k8sinternal/client_test.go
```diff
@@ -1,12 +1,15 @@
-package k8sinternal
+package k8sinternal_test
```
alexhope61, line 1: Is this the normal convention that the package test have a different name than the package itself? I've not done this before and I'm wondering if I should have been...

alexhope61 (approved):

internal/k8sinternal/client_test.go
rxbchen: I remember is something to do with blackbox/whitebox testing 💭. I think when we call this a different name, we would have to import `k8sinternal` in order to get access to the rest of the code.

rxbchen (approved):
Looks good!

internal/k8sinternal/client_test.go
genevieveluyt: Exactly what Brian said! It depends what you want to test.

If you use the same package as the code you're testing (whitebox), you have access to the private methods in the package and can test those.

If you use the special `_test` package (normally Go won't let you have multiple package names in the same directory, so this test package is a special case), you can test your code as if you were a consumer of the package (blackbox).

In this case we're not testing any private methods so using the "test" package is kind of a cheat to get out of import cycle problems. Alternatively we could move the functions that cause the import cycle into a new package and import that from both `k8sinternal` and `test`.

PR #428, opened 2022-05-24
✨ Adds "deprecated apis" auditor

### LINKED ISSUE #408: Deprecated API auditor
##### ISSUE TYPE
 - [ ] Bug Report
 - [x] Feature Idea

##### SUMMARY
The idea is to create a new auditor that'll check for deprecated api versions of resources. Suggested by @jerr

jerr:
##### Description
Fixes #408

Adds a new command `deprecatedapis` to find deprecated APIs of resources. It reports for all resources defined with a deprecated API the version since when it is deprecated  the deprecated version, the version where is removed and suggests the recommended API.

Here's a sample result:

```
$ kubeaudit deprecatedapis -f "auditors/deprecatedapis/fixtures/cronjob.yml"

  apiVersion: batch/v1beta1
  kind: CronJob
  metadata:
    name: hello
-- [warning] DeprecatedAPIUsed
   Message: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
   Metadata:
      DeprecatedMajor: 1
      DeprecatedMinor: 21
      RemovedMajor: 1
      RemovedMinor: 25
      ReplacementKind: CronJob
      ReplacementGroup: batch/v1
```

This PR also brings the possibility to audit any type of resource using the [dynamic client](https://github.com/kubernetes/client-go/tree/master/examples/dynamic-create-update-delete-deployment#create-update--delete-deployment-with-the-dynamic-package).   

##### Type of change

- [x] New feature :sparkles:

- [x] Automated tests 
- [x] Manual tests (local mode and manifest files)

##### Checklist:

### DIFF
--- a/docs/auditors/deprecatedapis.md
+++ b/docs/auditors/deprecatedapis.md
@@ -1,0 +1,96 @@
+# Kubernetes Deprecated API Auditor (deprecatedapis)
+
+Finds any resource defined with adeprecated API version.
+
+## General Usage
+
+```
+kubeaudit deprecatedapis [flags]
+```
+
+### Flags
+| Short   | Long                   | Description                                   | Default             |
+| :------ | :--------------------- | :-------------------------------------------- | :------------------ |
+|         | --current-k8s-version  | Kubernetes current version                    |                     |
+|         | --targeted-k8s-version | Kubernetes version to migrate to              |                     |
+
+
+Also see [Global Flags](/README.md#global-flags)
+
+## Examples
+
+The `deprecatedapis` auditor allows to find the deprecated APIs in use and indicates the versions where they will be removed and replacement APIs.
+```
+$ kubeaudit deprecatedapis -f "auditors/deprecatedapis/fixtures/cronjob.yml"
+
+---------------- Results for ---------------
+
+  apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: hello
+
+--------------------------------------------
+
+-- [warning] DeprecatedAPIUsed
+   Message: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
+   Metadata:
+      DeprecatedMajor: 1
+      DeprecatedMinor: 21
+      RemovedMajor: 1
+      RemovedMinor: 25
+      ReplacementKind: CronJob
+      ReplacementGroup: batch/v1
+```
+
+The `deprecatedapis` auditor can be used `--targeted-k8s-version` flag. If the API is not yet deprecated for this version the auditor will produce an `info` otherwise a `warning`.
+```
+$ kubeaudit deprecatedapis --current-k8s-version 1.20  -f "auditors/deprecatedapis/fixtures/cronjob.yml"
+
+---------------- Results for ---------------
+
+  apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: hello
+
+--------------------------------------------
+
+-- [info] DeprecatedAPIUsed
+   Message: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
+   Metadata:
+      DeprecatedMajor: 1
+      DeprecatedMinor: 21
+      RemovedMajor: 1
+      RemovedMinor: 25
+      ReplacementKind: CronJob
+      ReplacementGroup: batch/v1
+```
+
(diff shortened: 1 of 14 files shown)


jerr:
👆 rebased on the `mani` branch

internal/k8sinternal/client.go
```diff
@@ -65,6 +65,9 @@ func NewKubeClientLocal(configPath string) (KubeClient, error) {
 		return nil, err
 	}
 
+	// Ignore warnings from kubeclient as they are expected to be reported by the deprecatedapi auditor.
+	kubeconfig.WarningHandler = rest.NoWarnings{}
+
```
jerr, line 70: This  will hide warning messages such as the following:
```
W0608 12:04:51.010357   30221 warnings.go:70] v1 ComponentStatus is deprecated in v1.19+
W0608 12:04:54.355517   30221 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

Kubeaudit has to use the deprecated APIs to include them in the deprecatedapis reports.  
See https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

docs/auditors/deprecatedapis.md
```diff
@@ -0,0 +1,96 @@
...
+   Metadata:
+      DeprecatedMajor: 1
+      DeprecatedMinor: 21
+      RemovedMajor: 1
+      RemovedMinor: 25
+      ReplacementKind: CronJob
+      ReplacementGroup: batch/v1
+```
+
+The `deprecatedapis` auditor can be used `--targeted-k8s-version` flag. If the API is not yet deprecated for this version the auditor will produce an `info` otherwise a `warning`.
```
genevieveluyt, line 46:
```suggestion
The `deprecatedapis` auditor can be used with the `--targeted-k8s-version` flag. If the API is not yet deprecated for this version the auditor will produce an `info` otherwise a `warning`.
```

docs/auditors/deprecatedapis.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+kubeaudit deprecatedapis [flags]
+```
+
+### Flags
+| Short   | Long                   | Description                                   | Default             |
+| :------ | :--------------------- | :-------------------------------------------- | :------------------ |
+|         | --current-k8s-version  | Kubernetes current version                    |                     |
+|         | --targeted-k8s-version | Kubernetes version to migrate to              |                     |
```
genevieveluyt, line 15: I'm a bit unclear why there is a flag for both current and target version. Maybe I'm misunderstanding the use case, but to me it sounds easier to reason about if we had just the target version:

- `ERROR` if the resource API is removed in the target version
- `WARN` if the resource API is deprecated in the target version
- `INFO` if the resource API is not deprecated in the target version but will be deprecated in a future version

We could default to the current version in local and cluster mode since kubeaudit should be able to look that up automatically?

genevieveluyt (commented):
No major concerns though I would like to put this in the next release, so will approve / merge when the current release it out!

docs/auditors/deprecatedapis.md
jerr: If we specify a `current version` or this information is retrieved from the cluster without a targeted version we can have the following cases:
- `ERROR` if the resource API is removed in the current version
- `WARN` if the resource API is deprecated in the current version
- `INFO` if the resource API is not deprecated in the current version but will be deprecated in a future version

If we use a targeted version the message level takes precedence over the current version.

But maybe we can remove this option and only use the current version in the local or in cluster mode 🤔

genevieveluyt: Ah I see from your flag description

```
--targeted-k8s-version string   Kubernetes version to migrate to (eg 1.24)
```
The use case for this auditor is for doing Kubernetes version migrations?

Based on the description of the auditor "This command determines which resource is defined with a deprecated API version." I was thinking of "target version" as "check Kubernetes resources against this version", hence it didn't make sense to me to have both "current" and "target" versions. Did you have a specific workflow in mind where you would benefit from passing both current and target versions (eg. running against a live cluster as a one-off before doing a migration)? If so, I think we should document it (eg. as an example) to make the use of the flags more clear. Nothing (other than the target flag description) says anything about migrations.

jerr: Many APIs have been declared deprecated since the 1.16 but are removed from the version 1.22.  If the current version is before the 1.16 we can keep the API as it is but  if it is not the case, it is better to migrate the API as soon as possible. That is why a warning is generated from the version 1.16.

But as currently `kubeaudit` does not retrieve  the current version from the cluster we can remove this option in a first time.

genevieveluyt (dismissed):

genevieveluyt (approved):

PR #433, opened 2022-05-30
:sparkles: Use dynamic kubeclient

jerr:
##### Description

To be able to audit any type of resource a [dynamic client](https://github.com/kubernetes/client-go/blob/cf13620092b8545132ef4be034458e8cc29af5d6/dynamic/interface.go#L17) is used instead of a [clientset](https://github.com/kubernetes/client-go/blob/cf13620092b8545132ef4be034458e8cc29af5d6/kubernetes/clientset.go#L19).

##### Type of change

- [ ] Bug fix :bug:
- [x] New feature :sparkles:

All tests related to auditing resources from a cluster still work.
This features is used by the "deprecated apis" auditor (#428) to check all resources of a cluster.

##### Checklist:


internal/k8sinternal/client.go
```diff
@@ -89,288 +104,108 @@ type ClientOptions struct {
...
+				continue
+			}
+			for _, apiresource := range list.APIResources {
+				if len(apiresource.Verbs) == 0 {
+					continue
+				}
+				gvr := schema.GroupVersionResource{Group: gv.Group, Version: gv.Version, Resource: apiresource.Name}
+
+				// Namespace has to be included as a resource to audit if it is specified.
+				if apiresource.Name == "namespaces" && options.Namespace != "" {
```
genevieveluyt, line 146: If I'm understanding correctly, we need to treat namespaces differently because it's a `Get` instead of a `List`? We need the namespace resources for the [netpol audit](https://github.com/Shopify/kubeaudit/blob/main/auditors/netpols/netpols.go#L43) so I think we always want to include the namespace resource. If `options.Namespace != ""` then we only want to include the namespace resource where the namespace name is equal to `options.Namespace`, but it looks like `kc.dynamicClient.Resource(gvr).Get(context.Background(), options.Namespace, metav1.GetOptions{})` takes care of that filtering?
```suggestion
				// Namespace has to be included as a resource to audit if it is specified.
				if apiresource.Name == "namespaces" {
```

internal/k8sinternal/runtime_test.go
```diff
@@ -80,25 +77,22 @@ func TestGetObjectMeta(t *testing.T) {
 	deployment := k8s.NewDeployment()
 	deployment.ObjectMeta = objectMeta
 	deployment.Spec.Template.ObjectMeta = podObjectMeta
-	assert.Equal(objectMeta, *k8s.GetObjectMeta(deployment))
-	assert.Equal(podObjectMeta, *k8s.GetPodObjectMeta(deployment))
+	assert.Equal(&objectMeta, k8s.GetObjectMeta(deployment))
```
genevieveluyt, line 80: Out of curiosity: this looks functionally equivalent, is there an advantage to doing it this way?

genevieveluyt (commented):
Amazing! This is what I always wanted and never knew existed! Is there documentation that you use for discovery and dyanmic client or just the godoc? I have googled so many times for "how to unmarshal kubernetes objects from yaml in go" and never found it!!

My main concern is that we need the namespace resources, the rest are nits.

internal/k8sinternal/client.go
jerr: If the `options.Namespace` is not specified we want to retrieve all namespaces that is why we need to use the [`List`](https://github.com/Shopify/kubeaudit/blob/6e25d12b976b3ac45d0d0ce31866b18a243d233b/internal/k8sinternal/client.go#L155) function instead of the [`Get`](https://github.com/Shopify/kubeaudit/blob/6e25d12b976b3ac45d0d0ce31866b18a243d233b/internal/k8sinternal/client.go#L147) one. 

With the `options.Namespace != ""` test we will use the `Get` will return the namespace according to the option otherwise the `List` will be used to return all namespaces. 

If we use only the `List` that will return an empty unstructuredList  if a namespace is specified:
https://github.com/Shopify/kubeaudit/blob/6e25d12b976b3ac45d0d0ce31866b18a243d233b/internal/k8sinternal/client.go#L155

genevieveluyt: Ohh ok that makes sense. Neat, thank you for the clarification!

genevieveluyt (approved):

internal/k8sinternal/runtime_test.go
genevieveluyt: Gotcha, missed the type switch and thought it was some go convention I didn't know about 😄

PR #440, opened 2022-06-09
🐛 Ignore groups not served  by the cluster

jerr:
##### Description

If a group is not available on a server the `discovery.ServerPreferredResources` function returns  the list of preferred resources but also an [`ErrGroupDiscoveryFailed`](https://github.com/kubernetes/client-go/blob/v0.24.1/discovery/discovery_client.go#L331) error with the groups not available. 

This error can be ignored  because the unavailable resource types  should not be audited.

#
This is an example of a cluster with the `external.metrics.k8s.io` group not available:
```
$ kubectl  get apiservice v1beta1.external.metrics.k8s.io 
NAME                              SERVICE                    AVAILABLE                      AGE
v1beta1.external.metrics.k8s.io   my-custom-metrics-server   False (FailedDiscoveryCheck)   22d
```
```
$ kubectl  get --raw /apis/external.metrics.k8s.io/v1beta1
Error from server (ServiceUnavailable): the server is currently unable to handle the request
```
If we have a cluster in this state no resources are audited because an error is returned by ` discovery.ServerPreferredResources`
https://github.com/Shopify/kubeaudit/blob/6a30cc08c46cee8922f451b0449923b7a3ec761b/internal/k8sinternal/client.go#L129-L130

#
##### Type of change

- [x] Bug fix :bug:

- [x] Automated tests still work
- [x] Manual tests: The `kubeaudit all` command was tested on a cluster with an unavailable  `external.metrics.k8s.io` group. Without this fix no reports are displayed,  with the fix the reports are provided as expected.
  
##### Checklist:

### DIFF
--- a/internal/k8sinternal/client.go
+++ b/internal/k8sinternal/client.go
@@ -212,7 +212,8 @@ func (kc kubeClient) GetKubernetesVersion() (*version.Info, error) {
 func (kc kubeClient) ServerPreferredResources() ([]*metav1.APIResourceList, error) {
 	list, err := discovery.ServerPreferredResources(kc.discoveryClient)
 	// If a group is not served by the cluster the resources of this group will not be audited.
-	if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); ok {
+	var e *discovery.ErrGroupDiscoveryFailed
+	if errors.As(err, &e) {
 		return list, nil
 	}
 	return list, err
--- a/internal/k8sinternal/client.go
+++ b/internal/k8sinternal/client.go
@@ -106,7 +106,7 @@ type ClientOptions struct {
 
 type KubeClient interface {
 	// GetAllResources gets all supported resources from the cluster
-	GetAllResources(options ClientOptions) []k8s.Resource
+	GetAllResources(options ClientOptions) ([]k8s.Resource, error)
 	// GetKubernetesVersion returns the kubernetes client version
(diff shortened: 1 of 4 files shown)


internal/k8sinternal/client.go
```diff
@@ -207,5 +210,10 @@ func (kc kubeClient) GetKubernetesVersion() (*version.Info, error) {
 
 // ServerPreferredResources returns the supported resources with the version preferred by the server.
 func (kc kubeClient) ServerPreferredResources() ([]*metav1.APIResourceList, error) {
-	return discovery.ServerPreferredResources(kc.discoveryClient)
+	list, err := discovery.ServerPreferredResources(kc.discoveryClient)
+	// If a group is not served by the cluster the resources of this group will not be audited.
+	if _, ok := err.(*discovery.ErrGroupDiscoveryFailed); ok {
```
genevieveluyt, line 215: Can we use `errors.Is`?

genevieveluyt (approved):
Thanks! Just an optional nit

internal/k8sinternal/client.go
jerr: > Can we use `errors.Is`?

`errors.Is` needs an error instance right? We want to check its type here.

genevieveluyt: `errors.As` then? 😄

genevieveluyt (approved):

PR #469, opened 2022-08-18
refactors sarif test

dani-santos-code:
reference: https://github.com/Shopify/kubeaudit/pull/453/files#r938176410

this PR:

- exports the results field in the Report struct:

```go
type Report struct {
	Results []Result
}
```
- renames results function to GetResults:
```go
func (r *Report) GetResults() []Result
```


genevieveluyt (approved):
I added some commits to:

1. Add a `NewReport` function so we don't have to make any breaking changes
2. Make `workloadResult` exported so you can use it in your test

Lmk what you think. The tests themselves look great!

PR #486, opened 2022-10-05
adds metadata to sarif result

dani-santos-code:
This is to add the Metadata field/info to the sarif. We currently display this info on the CLI. but it was missing on the sarif. Here's an example of what it will look like on Github UI:

Screenshot 2023-01-23 at 5 45 14 PM

### DIFF
--- a/internal/sarif/sarif.go
+++ b/internal/sarif/sarif.go
@@ -45,7 +45,7 @@ func Create(kubeauditReport *kubeaudit.Report) (*sarif.Report, error) {
 
 		metadata, jsonErr := json.Marshal(formattedMap)
 
-		if err != nil {
+		if jsonErr != nil {
 			metadata = []byte(jsonErr.Error())
 		}
 
--- a/internal/sarif/sarif.go
+++ b/internal/sarif/sarif.go
@@ -40,18 +40,23 @@ func Create(kubeauditReport *kubeaudit.Report) (*sarif.Report, error) {
 		formattedMap := make(map[string]string)
 
 		for k, v := range result.Metadata {
-			formattedMap[k] = "`" + v + "`"
+			formattedMap[k] = v
 		}
 
 		metadata, jsonErr := json.Marshal(formattedMap)
-
 		if jsonErr != nil {
 			metadata = []byte(jsonErr.Error())
 		}
 
+		var metadataTxt string
+
+		if len(formattedMap) > 0 {
+			metadataTxt = fmt.Sprintf("Metadata: %s\n", string(metadata))
+		}
+
 		docsURL := "https://github.com/Shopify/kubeaudit/blob/main/docs/auditors/" + auditor + ".md"
 
-		helpText := fmt.Sprintf("Type: kubernetes\nAuditor Docs: To find out more about the issue and how to fix it, follow [this link](%s)\nDescription: %s\nMetadata: %s\n\n Note: These audit results are generated with `kubeaudit`, a command line tool and a Go package that checks for potential security concerns in kubernetes manifest specs. You can read more about it at https://github.com/Shopify/kubeaudit ", docsURL, allAuditors[auditor], string(metadata))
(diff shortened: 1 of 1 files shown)


internal/sarif/sarif.go
```diff
@@ -36,12 +37,24 @@ func Create(kubeauditReport *kubeaudit.Report) (*sarif.Report, error) {
 
 		auditor := strings.ToLower(result.Auditor)
 
+		formattedMap := make(map[string]string)
+
+		for k, v := range result.Metadata {
+			formattedMap[k] = "`" + v + "`"
```
thepwagner, line 43: Question: are you sure that `v` won't contain a backquote? That could break your formatting. Example:

`meow ` meow`


If you are formatting for GitHub markdown, I think double-backticks may work - ``meow ` meow``
If you are formatting for humans, maybe ``strings.Replace(v, "`", "'")`` to sanitize? Humans will probably still get the gist.

If your formatting for HTML, maybe ``strings.Replace(v, "`", "&#96;")``? That is weird - it only came up because https://github.com/github/markup/issues/363

thepwagner (dismissed):
I think `err` vs `jsonError` is worth fixing.

Other comments stem from me not fully understanding what `Metadata` will contain, and so getting paranoid about edge cases. Those are safe to ignore!

internal/sarif/sarif.go
```diff
@@ -36,12 +37,24 @@ func Create(kubeauditReport *kubeaudit.Report) (*sarif.Report, error) {
...
+		}
+
+		metadata, jsonErr := json.Marshal(formattedMap)
+
+		if err != nil {
```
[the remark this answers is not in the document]
dani-santos-code: good catch! meant to check for `jsonErr` indeed

internal/sarif/sarif.go
dani-santos-code: usually, `v` is the name of the resource. I don't think it'll contain a backquote, but if anything changes down the line, we might run into issues. great point!

Yeah, I'm thinking Github markdown for sure. Considering the backticks for extra emphasis, but in the end we don't really need it? If I come to think of it, "`web`" looks weird...

internal/sarif/sarif.go
```diff
@@ -36,12 +37,24 @@ func Create(kubeauditReport *kubeaudit.Report) (*sarif.Report, error) {
...
+			metadata = []byte(jsonErr.Error())
+		}
+
 		docsURL := "https://github.com/Shopify/kubeaudit/blob/main/docs/auditors/" + auditor + ".md"
 
-		helpText := fmt.Sprintf("Type: kubernetes\nAuditor Docs: To find out more about the issue and how to fix it, follow [this link](%s)\nDescription: %s\n\n Note: These audit results are generated with `kubeaudit`, a command line tool and a Go package that checks for potential security concerns in kubernetes manifest specs. You can read more about it at https://github.com/Shopify/kubeaudit ", docsURL, allAuditors[auditor])
+		helpText := fmt.Sprintf("Type: kubernetes\nAuditor Docs: To find out more about the issue and how to fix it, follow [this link](%s)\nDescription: %s\nMetadata: %s\n\n Note: These audit results are generated with `kubeaudit`, a command line tool and a Go package that checks for potential security concerns in kubernetes manifest specs. You can read more about it at https://github.com/Shopify/kubeaudit ", docsURL, allAuditors[auditor], string(metadata))
```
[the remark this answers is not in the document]
dani-santos-code: https://github.com/Shopify/kubeaudit/pull/486/commits/c5f32343e68ef9d11af9c4cde1f3db549dded508

examples of what it looks like right now can be found in the test repo

internal/sarif/sarif.go
dani-santos-code: Screenshot 2023-01-23 at 5 45 14 PM

dani-santos-code: ^ opted for this format

thepwagner (approved):
No blockers, just an optimization idea.

Issue #538, opened 2023-03-14
docker hub free team account being sunset


dani-santos-code:
We won't be able to host our image for free anymore

We'll be looking into:

- [moving to GitHub packages ](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)
- use [Github releaser action](https://github.com/goreleaser/goreleaser-action) for publishing

PR #539, opened 2023-03-14
go releaser action

dani-santos-code:
##### Description

This PR adds a releaser action that will create a draft release on Github and publish the [library to the github registry](https://github.com/orgs/Shopify/packages?repo_name=kubeaudit). Notice: this has been used to publish a package and make a release already

After talking to Genevieve, we've decided not to keep a separate `CHANGELOG.md` file. We can either go with the body generated by go releaser or click this button to automatically tag people in contributions

Screenshot 2023-03-27 at 9 44 12 AM

I've tagged this commit a couple of times to test the plan

The plan is:

after this gets approved/merged, we'll re-tag this commit on main and trigger the action again and finally make a release! 

Closes https://github.com/Shopify/kubeaudit/issues/538

example draft release: https://github.com/Shopify/kubeaudit/releases/tag/untagged-565c4cad9b0e59740674

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -25,14 +25,16 @@ git push origin v0.11.6
 ```
 
 
-6. Once you push the tag, the release Github action will be triggered and generate a draft release in Github, allowing you to double check it and make changes to the Changelog. Find the [draft release](https://github.com/Shopify/kubeaudit/releases) and make sure there are no commits to main since the release.
+5. Once you push the tag, the release Github action will be triggered and generate a draft release in Github, allowing you to double check it and make changes to the Changelog. Find the [draft release](https://github.com/Shopify/kubeaudit/releases) and make sure there are no commits to main since the release.
 
 > If there are commits to main since the release, this may mean you didn't make the tag on main or your main is out of date.
 
-7. Click `Edit` on the right of the draft release and tidy up the Changelog if necessary. We like to add thank you's to external contributors, for example:
+6. Click `Edit` on the right of the draft release and tidy up the Changelog if necessary. We like to add thank you's to external contributors, for example:
 
 ```
 202e355 Fixed code quality issues using DeepSource (#315) - Thank you @withshubh for the contribution!
 ```
 
-8. Click on `Publish release` at the bottom.
+Optionally, you can click on "Generate release notes", which adds Markdown for all the merged pull requests from the diff and contributors of the release.
+
+7. Click on `Publish release` at the bottom.


CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
...
+
+# 0.22.0
+
+* support for metadata info on sarif result
+* Makefile fix for go mod tidy
+* test without Kind by default
+* deprecate kubernetes.io in override labels
+
+
+# 0.21.0
```
dani-santos-code, line 12: this is not being used to automatically populate the draft release. However, it'd be nice to keep as a way to document things.  Should we keep it or not?

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      -
+        name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
```
thepwagner, line 27: Idea: if you track the semver tag in a comment, Dependabot should update both!
[example](https://github.com/Shopify/hansel/blob/543acc5e504feea44b8b1268e2bc35d5fa75a8f1/.github/workflows/release.yaml#L17) / [example PR](https://github.com/Shopify/hansel/pull/111).

I really dig it: you get the immutability of pinning a tag, without losing the visibility that the tag provides.

.goreleaser.yml
```diff
@@ -5,6 +5,15 @@ release:
...
   name_template: "{{.ProjectName}}-v{{.Version}}"
+dockers:
+- dockerfile: goreleaser.Dockerfile
+  goos: linux
+  goarch: amd64
+  goarm: ''
+  image_templates:
+  - "ghcr.io/shopify/kubeaudit:latest"
+  - "ghcr.io/shopify/kubeaudit:{{ .Tag }}"
+  - "ghcr.io/shopify/kubeaudit:v{{ .Major }}.{{ .Minor }}"
```
thepwagner, line 16: Noice: this pattern is cool!

thepwagner (approved):
This is dope! By building the release artifacts in an ephemeral runner, we are on our way to SLSA town 🍅🌶️ !

In terms of process, I tend to drive things more from the [Create a new release](https://github.com/Shopify/kubeaudit/releases/new)
* I use the GitHub tooling to generate a changelog/release notes
* Pushing the "publish release" button is what creates the tag that starts the workflow to build the artifacts
* That version is pushed in to my build (via `-X github.com/Shopify/kubeaudit/cmd.Version={{.Version}}` like this build is doing). I dig how this avoids maintaining a `VERSION` file.  I don't see the `cmd.{Version,Commit,BuildDate]` variables that are being [embedded via `-X` ](https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications) actually being used.

It's cool because I click the button and the robots take over from there. It's not cool because sometimes things break, and you either have to undo+redo the release or fail forward quickly.

---

The other can of worms to consider is signing! By building your release artifacts in GitHub Actions, you can use the Actions worker's identity to get a short-lived signing key from [fulcio!](https://github.com/sigstore/fulcio).

You may have to add `sigstore/cosign-installer` to the worfklow, then you can add a `docker_signs` section to the goreleaser configuration [example](https://github.com/Shopify/hansel/blob/543acc5e504feea44b8b1268e2bc35d5fa75a8f1/.goreleaser.yaml#L65-L72)

With that, the whole internet can verify the artifact they are receiving was built by Actions executing your workflow:
```
# Assumes cosign >= 2.0.0
$ cosign verify --certificate-oidc-issuer https://token.actions.githubusercontent.com --certificate-identity https://github.com/Shopify/hansel/.github/workflows/release.yaml@refs/tags/v0.0.9 --certificate-github-workflow-repository Shopify/hansel ghcr.io/shopify/hansel:0.0.9
Verification for ghcr.io/shopify/hansel:0.0.9 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates
```

You can use the same technique to sign other release artifacts too! https://goreleaser.com/customization/sign/#signing-with-cosign In hansel, I use that and the `.release.extra_files[]` configuration to sign the blobs and the capture the certificate/signature as part of the uploaded release:

```
$ cosign verify-blob --certificate-oidc-issuer https://token.actions.githubusercontent.com --certificate-identity https://github.com/Shopify/hansel/.github/workflows/release.yaml@refs/tags/v0.0.9 --certificate ./hansel_0.0.9_darwin_amd64-keyless.pem --signature ./hansel_0.0.9_darwin_amd64-keyless.sig hansel
Verified OK
```


Sorry for all the exclamation points, I just dig this stuff.

.goreleaser.yml
dani-santos-code: this is the pattern that we had before, but were pointing to docker hub registries instead

.github/workflows/release.yml
dani-santos-code: oh, neat! will add it! thank you!

dani-santos-code: will be addressed in a separate PR as per: https://github.com/Shopify/kubeaudit/issues/547

Issue #593, opened 2024-08-19
Deprecate kubeaudit


dani-santos-code:
### Context

We should consider deprecating `kubeaudit` since Shopify's team hasn't been able to keep up with maintenance and cannot currently be efficient stewards.

We believe [kube-bench](https://github.com/aquasecurity/kube-bench) is a good successor. 

### Tasks

- [x] update README mentioning that the project isn't maintained and mention an alternative (`kube-bench`). Make changes to Issues and PR templates
- [x] update the code to add warnings (help +  running with flags)
- [x] make a new and last release
- [x] set up an announcement in [Discussions](https://github.com/Shopify/kubeaudit/discussions/new?category=announcements)
- [ ] finally archive the repo by the due date agreed on (October 2024)

hrmcardle0:
Is there a large difference between the two, is it worth it for team to takeover ownership of this? I have not used it enough yet to determine this.

dani-santos-code:
hi @hrmcardle0. which team are you talking about? 🤔

PR #595, opened 2024-08-21
Update version to v0.22.2

Shariatj:
Releasing a new version of Kubeaudit for the depracation notice.
Related issue:https://github.com/Shopify/kubeaudit/issues/593#issue-2473964819


jacobmichels:
In the interest of maintaining a clean commit history, would you mind squashing these commits into one?

jacobmichels:
Does bumping this version automatically trigger a GH release once merged to main? Or is there a separate process for that

Shariatj:
> Does bumping this version automatically trigger a GH release once merged to main? Or is there a separate process for that

No we have to manually relaese it by creating a tag, ref: https://github.com/Shopify/kubeaudit/blob/main/docs/release.md

cmd/commands/VERSION
```diff
@@ -1 +1 @@
-0.22.1
+0.23.0
```
dani-santos-code: q: shouldn't this be a patch bump? we're not introducing new functionality and it's backwards compatible

Shariatj: Since this isn't a bug fix, adding a deprecation notice (considered a minor update without breaking functionality), I thought the next version should be v0.23.0. Could be changed to a patch bump too

Shariatj: @dani-santos-code changed it into patch bump now.

dani-santos-code: yeah, it's not a bug fix. It seems debatable, though. 😬 Found some discussions on StackOverflow around this and I don't see a clear consensus:
- https://stackoverflow.com/questions/52373074/how-to-semver-versioning-documentation-update
- https://docs.spryker.com/docs/dg/dev/architecture/module-api/semantic-versioning-major-vs.-minor-vs.-patch-release.html#what-is-a-major-release

dani-santos-code: This is not a blocker at all

dani-santos-code (approved):
lgtm! can we update the title of the PR? thanks!

